PR #2080, opened 2013-04-09
Climate zones and island fertility [NEED TO BE DONE]

ChrisOelmueller:
2013_3-fertility
Able to:
* display resource icons

Currently assumes a 0/1 fertility system (either resource A can exist on island, or it cannot).
There are no stepdowns like "half efficiency" in the [current design] (http://wiki.unknown-horizons.org/w/Fertility) and I'm not sure they add that much gameplay value, as opposed to interface burden and code complexity (which they definitely do add).

The current mapping for resources against buildings indicates that if an island is infertile w.r.t. a particular resource, the corresponding building in MAPPING should be impossible to build in settlements on that island.
How to indicate that from our build menus, message widget, buildability highlights, or a combination thereof, is up to you.

Yes, this is gathering resource<->building relations in code instead of our database. Sorry about that. Will look into wiring those in later.

Edit by nihathrael:

Open tasks:
- [x] Fair resource distribution among islands
- [x] Save/load the resources for each island
- [x] Always display resource menu, even when not having a settlement on the island
- [x] Finalize resource distribution on fertility zones
- [ ] GUI that shows exploration button and progress in the ship's UI
- [ ] Only show resource menu when island is explored
- [ ] Finalize mapping of resource<->building
- [ ] AI has to be able to handle climate zones


LinuxDonald:
Any news here? and can it be merged into master?

LinuxDonald:
Closing it for now. Created an issue for it: https://github.com/unknown-horizons/unknown-horizons/issues/2407

Issue #636, opened 2012-01-23
Implement mountains


ChrisOelmueller:
Mountains at the moment basically are tiles placed above ground tiles where nothing (besides mines) can be built. The behaviour is a bit similar to the clay ground where clay pits (and only clay pits) can be placed on.

Since we only have few and rather old graphics, sizes are not fixed. We still need to discuss advanced mechanics for the following questions:
* do we want mountains that can tile like streets? they would be of size 2 x ? then and allow mining along their edge
* mining concepts: how many mines per mountain? what exactly is *one* mountain in the above tiling concept?
* are all mountains deposits of a resource? => do we see stonemasons as a kind of mine, too?

[> Link to originally reported Trac ticket <] (http://trac.unknown-horizons.org/t/ticket/636)

LinuxDonald:
Reopening see PR: https://github.com/unknown-horizons/unknown-horizons/pull/2202
Need to be fixed/making it done.

PR #2202, opened 2013-11-19
Mountains [NEED TO BE FIXED]

DanielStephens:
**Please do not merge this yet.**

This pull request is to keep track of what still need to be done to be able to complete #636 and hopefully get some help from those more knowledgable on how to solve some issues.

- [ ] I am yet to test this code with the AI but if the AI does not build mines properly then I will need help on AI coding.
- [ ] The graphics for the mine are placed on top of the ground tiles. I can use transparency to fake them being behind mountains sometimes. but for some use cases this does not work:
 - Near corners
 - If we allow players to build single height mountains as well as double height then graphics for the mine will be very wrong

My Ideas for sorting this currently are:
- Fake the mine being a building completely and have it drawn in the ground layer (very hacky)
- Allow different graphics to be used by the mine in different situations (pretty hacky)
- Enable some kind of z-ordering between layers (I'm guessing this would hit performance?)
- Fake 'mountain buildings' and match these to the tiles beneath so that the z-ordering works on the building layer, and mines are then drawn behind the mountain buildings

I will have a similar problem to the mines when showing deposit graphics, though probably less so.


LinuxDonald:
Any news here?

LinuxDonald:
Closing it for now: https://github.com/unknown-horizons/unknown-horizons/issues/636

PR #2272, opened 2014-03-15
Fix cursor in editor [NEED TO BE FIXED]

ThePawnBreak:
I noticed that when editing a map and the cursor is set to a tile, it doesn't change back to an arrow when hovering over the GUI. I made it do that.
There's a small problem in the way I implemented it: you must always have a tile selected, which means that right click doesn't change back to arrow cursor anymore. I don't know how to fix this, since I can't find an event for right click. Please let me know what you want me to do about this.

### DIFF
--- a/horizons/editor/gui.py
+++ b/horizons/editor/gui.py
@@ -205,6 +205,8 @@ class SettingsTab(TabInterface):
 		super(SettingsTab, self).__init__(widget=self.widget)
 
 		self._world_editor = world_editor
+		self.current_tile = 'sand'
+		self.ingame_gui = ingame_gui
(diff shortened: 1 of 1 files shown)


LinuxDonald:
Any update here?

ThePawnBreak:
I haven't worked on UH for more than a year, and I'm too lazy to set up an environment just to finish this PR. If you want to, feel free to open up another PR and use any part of my code that you find useful.

LinuxDonald:
Closing it and opened an issue for it: https://github.com/unknown-horizons/unknown-horizons/issues/2409

Issue #2199, opened 2013-11-18
Add ability to manually place resources onto map in editor


woox2k:
Title says pretty much everything, it would open up many new possibilities when designing a map.
Just to be clear i meant resources like forest, fish etc.

LinuxDonald:
PR for this need to be fixed: https://github.com/unknown-horizons/unknown-horizons/pull/2293

PR #2293, opened 2014-08-23
added ability to lay some resources for testing (WIP); fix for #2199 [NEED TO BE FIXED]

Zappaman:
This pull request attempts to fix https://github.com/unknown-horizons/unknown-horizons/issues/2199 .

What I did was add a new tool called ResourceTool, that acts very similarly to the TilelayingTool, only that it should work on buildings instead of tiles. I added two new labels to the gui in the editor, called 'fish_deposit' and 'tree', that when clicked, trigger the ResourceTool similarly to the way the tiles trigger the TilelayingTool.
I also added some necessary functions in intermediatemap.py and in the worldeditor.py to implement the functionality. Momentarily one can add fish deposits  by selecting the fish_deposit label (by clicking and dragging)

This is WIP. I have encountered some issues, and I would really appreciate some feedback/hints from people more knowledgeable than me:

1) I couldn't follow the same procedure to change the cursor image as with the tiles, when selecting the corresponding resource. I got a few fife errors. What images should I add for the cursor, and in what way? I think I saw that the tile images could be found procedurally, the images for the resources at this point could be hardcoded, but I'm pretty sure that is not the way to do it.

2) I was able to lay fish_deposits, but trees are proving to be a bit more problematique. In the Build constructor, when trying to build trees I get a `build_position.buildable == False` no matter what terrain I am laying it on (line 97 in building.py). Could anyone give me a hint on a workaround, please?

3) I need a way to validate the possibility of setting the resource on the map. Due to 2), I'm not sure that checking the tile in the full_map[coords] is enough to tell if that is possible.

I am open to any other remarks on the subject, feedback, hints, suggestions or other possible problems.

Thanks!

### DIFF
--- a/content/gui/xml/editor/editor_settings.xml
+++ b/content/gui/xml/editor/editor_settings.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
-<Container size="250,350" name="editor_settings">
-	<TabBG amount="4" />
+<Container size="250,400" name="editor_settings">
+	<TabBG amount="6" />
 
 	<Container position="0, 20" size="250, 350" max_size="250, 350">
 		<Label text="Select brush size:" name="headline_brush_size" position="25,55" />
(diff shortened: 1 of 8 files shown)


LinuxDonald:
Any news here?

LinuxDonald:
@Zappaman do you will work again on it?

LinuxDonald:
Closing it for now: https://github.com/unknown-horizons/unknown-horizons/issues/2199

Issue #1688, opened 2012-07-06
Routeconfig improvements


totycro:
This is a collection of usability problems of the routeconfig dialog:

* Route should change to enabled status whenever it is enableable, e.g. when the second city is added. It could be disabled when validity is destroyed. Tooltips in the ok- and cancel-button as well as the route enable button of course could give hints about that. Some highly visible coloring could be added, which is the reward for the player for configuring a route, the game assures you everything is fine by the coloring.

* The action could default to unload whenever a resource has been loaded in the route before the current station. 

* If a resource is filled to a slot, it should default to taking all the available res. This is more reasonable than taking none, which is never what the player wants

* Add "dump all" option for the end of routes. It could either be a special state that really tries to dump everything that happens to be in the ship, or it just automatically fills in the slots to dump everything that was loaded before.

qsorix:
Regarding the first suggestion, I'm not sure I would like the route to auto-enable as a player. I've just accidentally loaded some food onto my ship and due to over-production, it wasn't pleasant to get it off the ship. So if a route auto-enables while it isn't finish, it may give players some headaches.

However, I agree that it should be more intuitive the "enable" button has to be clicked. Highly visible coloring is, I guess, the way to go.

LinuxDonald:
See pr: https://github.com/unknown-horizons/unknown-horizons/pull/2307

AndyMender:
Savegame file where routeconfig-related bugs occur: https://www.file-upload.net/download-12442723/trade.sqlite.html (courtesy of @LinuxDonald).

This issue definitely needs to be addressed.

AndyMender:
@LinuxDonald, how far are we on this in your opinion? :)

LinuxDonald:
The routeconfig now looks really better than before. But what we need is player feedback on it.

PR #2307, opened 2015-01-10
Issue 1688 [NEED TO BE FIXED]

qsorix:
My attempt at 3 out of 4 problems mentioned in #1688. I left auto-enabling out as I don't agree with this suggestion.

The dialog looks like this after my changes
http://up.unknown-horizons.org/f/?id=692&n=traderoute.png

I'm happy to improve the code. If you don't like something, please do tell me.


qsorix:
@nihathrael, I want to cover new functionality with GUI tests, but they're failing at the moment. Seems like they've started after this change
https://github.com/unknown-horizons/unknown-horizons/commit/6967cd20bf11aee3fb2dd42ed828b24ed19d91e4

When tests are trying to show() the trade route configuration dialog, the window is not ready yet. Suggestions how to fix that are welcome.

ChrisOelmueller:
Thanks for working on these!
I'll have to admit that I was not a big fan of the "dump all" functionality around the time it first came up. It felt like throwing more complexity at an already complex system to solve some of the problems that complexity created in the first place.
So as a fair warning, that might bias my opinion on the current implementation a bit. I do like the representation with crates though.

So here's some collected and assuredly negative feedback, which is the open-source way of reviewing things. Of course there's lots of good stuff too but that already exists or is taken for granted, so why mention it!

The newly-introduced icon next to the arrows feels misplaced.
Proposal on how I think the presentation could be improved visually and conceptually:
A "dump row" could instead be created by offering a special resource in the resource selection dialog, much like the end result looks like right now.
Ideally the icon to select there would signal that it's a wildcard, e.g. the current crate icon with some kind of visual cue.
Right now the "amount" line seems to be unused, a short and concise label like `_("All")` (with appropriate i18n comment so that translations try not to overflow) might go a long way there.
Have you looked into (only for visuals) setting the sliders to their max values, and perhaps disabling or even removing them as well?

On paper, smart systems seem very nice, but it's very important that those systems still act in predictable ways. If the user can't figure out what is happening and why, they certainly won't like that feeling of either being outsmarted or struck by buggy behavior.
Again, if this works correctly and proposes unloading of a resource instead of loading if it was previously loaded on that route, I'd prefer just having players use that shortcut to their advantage instead of introducing all the complexity and interface burden that is "unload all".

qsorix:
Thanks for looking at the changes, @ChrisOelmueller!

> A "dump row" could instead be created by offering a special resource in the resource selection dialog, much like the end result looks like right now.

I considered that. One thing I didn't know how to resolve was that it is supposed to be a "dump _row_". For it being a row, player would need to pick the same dummy resource 3 times. Picking the resource once, should not change what's displayed in other slots. Of course it doesn't really require a full row. A single slot with unload-anything can trigger unloading the whole ship.

Finally, one slot of unload-anything could mean just that, unload one resource whatever that is. But then it doesn't work well because the dialog has 3 slots, and ship can carry e.g. 4 resources.

> a short and concise label like _("All") (...) setting the sliders to their max values, and perhaps disabling or even removing them as well?

Good ideas. I forgot that right now it is absolutely not obvious what this row is doing.

> If this works correctly and proposes unloading of a resource instead of loading if it was previously loaded on that route, I'd prefer just having players use that shortcut to their advantage instead of introducing all the complexity and interface burden that is "unload all".

I agree with the way you think, i.e. that the interface should be simple etc.. But playing a bit with this dump-all option available, I've found it very useful. This is because the dialog has much deeper problems actually, but I'll write about it later.

Assume you don't have unload-all. If you want to carry 4 resources from one island to another, and then 4 resources back, you need to have: 2 rows (3+1 slot) to load stuff from A. 2 rows to unload this stuff at B. 2 rows to load from B, and 2 rows to unload at A. 8 rows in total. Game allows 7.

With unload-all, you can do the same with 6 rows. Not a great improvement but at least lets you utilize the ships available today. It also let's you take the ship, whatever is on board, and assign it to the route. The resources on the ship will be dropped somewhere to free the slots.

I'll write about the auto-enabling in a separate comment.

qsorix:
Wall of text follows. Put your games on auto-pilot.

Auto-enabling is wrong because a partially configured route can load some stuff onto the ship, which will then be very hard to drop, e.g. when all your settlements have over production of wood.

I can think of several improvements here, but they all change gameplay. The question arises what is the goal here. Do we want players to spend time during game to keep an eye on all their ships and micro-tweak the settings all the time, or do we want the trading route to offer a lot of intelligent automation, so the player can set it up and leave running. There are way to make players' lives easier, but perhaps we don't want to as then the game becomes too easy.

Apart from auto-enabling, there are other things I don't like.
 - there are 4 (depends on the ship, but 4 is a good number for discussion) slots on the ship, but only 3 slots in the dialog., this makes you use more rows than you want
 - when arriving at a settlement where you want to unload and load, you need to add this settlement twice to your route
 - if your ship is lost, you loose the trading route
 - there's no indication at which entry of the route your ship is right now, which makes it harder to predict when will it arrive at some settlement
 - setting up two ships with the same route requires you to define the route twice

Once more I'll repeat that these would seem bad if you were writing utility software. But in the game, it may be OK to have something to fiddle with.

Some of the ideas I'm mentioning below assume that you have just a single ship and want to do everything with it... which may be unrealistic, and I agree we can decide it shouldn't be possible. Rules of he game.

The biggest improvement I've considered was defining trading routes and assigning them to ships independently, e.g. in the GUI on two different pages of the book. There would be no notion of a enabled route, the ship can start following it once it is assigned. It also helps when the ship is destroyed or you want more ships to do the job.

The other thing on my mind was to change the GUI to allow more slots in a row, allow more entries (vertical scroll), and to make each entry contain two rows of slots: loading and unloading slots. Then you would not need to add one warehouse twice, unless you want a route that has topology of a star not a cycle.

When route is enabled, the ship should not just start from the beginning. Based on resources on the ship, and orders in the route, and distance to each settlement, it could pick where to go first. E.g. if you're exchanging 2 resources between 2 islands (one resource in each direction), the ship can just go to the closer one at start. But I think it may be easier and better to highlight the route's entry that is being active at the moment, and add a "skip" button.

Currently the routes are oriented around physical act of loading/unloading stuff. You could instead provide an interface where you say: I want wooden boards to go from this island to this, I want bricks to travel from here to there, etc. And then the game could solve TSP to set up the route for you. It could even suggest the number of ships needed and assign ships automatically.

Routes are dealing with tons of resources. When playing, I've found this is not what I want. (Remember gameplay difficulty balancing!) I had two big settlements with production divided between the two. One was producing salt, the other food. I wanted to have over production from each island to go to the other. Instead of orders like "load 30t of salt" I wanted to do what you can when trading. "Load everything above 25t". Ideally, you want something like "Load salt. Take as much, as the other settlement needs in the time this ship travels both ways. Don't take more than this settlement requires to survive that time". Not taking too much is important, because eventually the other settlement gets full deposits and the ship cannot unload.

Another improvement: raise a message when ship cannot unload. And add a button to sink the resources on board. Otherwise dealing with over production is hard, you need to stop production chains and wait for the deposits to drop so the ship can unload freeing slots.

Another improvement: when ship has been waiting for considerable amount of time because of trade routes's options  "wait to unload" or "wait to load", raise a message. It can also be triggered sooner if the island doesn't have the resource, is not producing it, and is not buying it.

Other types of trading routes I've wanted are many-to-one or one-to-many. I played a game where islands were tiny. I had one island with habitants and some basic production (lumberjack, fisherman), the luxury goods like tabacco and alcohol came from other islands. I wanted a route to pick tabacco from 3 islands and carry it to my home island. Problems? If 3 islands produce more than ship's capacity and you create a simple cycle, one island cannot be fully served. If 3 producing islands are in one corner of the map, and the home is in the other, going 3 times back and forth is inefficient. You may want to cycle which production is served first:
 - take from A, take from B, take from C,
 - carry to Home,
 - take from B, take from C, take from A,
 - carry to Home,
 - then take from C take from A, take from B,
 - carry to Home,
 - repeat.

A similar case exists for unloading. In this case, I wanted to have percentage based unload. There's small demand for some goods, e.g. medical herbs. You can produce them on one small island, and transfer to 3 inhabited islands. You want
 - take all from Production,
 - unload 33% at Settlement A,
 - unload 33% at Settlement B,
 - unload 34% at Settlement C,

Today you can say take 30 from production, unload 10 at each settlement. But then to have a fair distribution you need to mark "wait for load". Then it is hard to say "take all".

One more thing I would want is an option to wait per slot, not per route. I may want to wait to load salt, but don't care if 1t of medical herbs is not available yet. Also, I may want to transfer salt to my settlement, and wait for unload, as I don't wan to waste salt. After my settlement takes their part, the ship can try and go sell something to other players. Waiting there is not wanted.

Oh, and it would be great if the ship would skip route entries that make no sense right now. E.g. if the ship is empty, there's no point to sail somewhere to unload.

Let's stop with ideas here but surely transfers and trading offer so much complexity, lots more can be said.

I've been think about the GUI to do some of this. So first, a separate screen to setup the route, then a separate screen to show assigned route in action on a ship.

The editing screen would need more space, because there are more slots, more rows, and for each slot you have a "wait" icon. Plus you may want to have more sliders, e.g. one to set minimum amount to leave on the island, the other to take how much to take. And maybe a button to switch between amounts and percentage.

The view of an active route can be very similar to what's now, but there'll be only one row per settlement and a summary. The row can be highlighted and the minimap can show the whole path, also highlighting the active section.

I'm open for brainstorming. Also the one that summarizes all this in: what's implemented now is enough. I honestly don't think implementing all of this is even remotely close to being a good idea. In a game where you need to manually place roads, it would be weird to have a fully automated trading system.

qsorix:
Changes updated after feedback. I didn't get the point about enumerate, I'm not sure how to use it my case in a way that won't be less readable than what's written now.

I've added a label "all" and removed the slider from unload-all entries. Otherwise the GUI stayed the same.

stubb:
Gave it a quick try. I recommend to disable the move up/move down buttons if it isn't possible. Like when there is only one entry yet and if there are multiple ones. The first only can be moved down and the last only be up.

qsorix:
@stubb , good idea as well, or instead of disabling the buttons, they could rotate entries from the very bottom to the very top and vice-versa. After all that's how the ship goes.

For the next week and a half I'm busy, but I may continue later.

LinuxDonald:
Any update here?

qsorix:
Hi there. Nope, sorry. Why do you ask? I doubt I'll have time to go back to this soon. Anybody is welcome to take over or start over or just drop this.

LinuxDonald:
https://github.com/unknown-horizons/unknown-horizons/issues/1688Ciolsing this for now

PR #2309, opened 2015-01-15
German translation - Little changes

Darksider3:
Had found some little wrong translations, and some stylistic ... "bad" things. Corrected, and commented, so other can understand why i've changed it. :smile:

### DIFF
--- a/po/terminology/de.po
+++ b/po/terminology/de.po
@@ -1071,7 +1071,7 @@ msgid "remove"
 msgstr "entfernen"
 
 msgid "renderer"
-msgstr "Renderer"
+msgstr "Kafiller" #or "Abdecker"..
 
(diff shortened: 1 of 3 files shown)


stubb:
Thanks for your work. Was there a reason to not use weblate for this? Normally we use weblate for all i18n stuff. Here is explained how you can use it with plain po files https://github.com/unknown-horizons/unknown-horizons/wiki/Translators#using-plain-po-files

Darksider3:
Sorry, changed back..

stubb:
Yep, please use weblate for your changes. More infos can be found here https://github.com/unknown-horizons/unknown-horizons/wiki/Translators. You should be able to upload your changed files there.

PR #2324, opened 2015-06-13
Correct grammar in Polish messages

macbre:


stubb:
Sorry for the late reply and thanks for your PR :+1: 
Normally we use weblate for all i18n stuff. Here is explained how you can use it with plain po files https://github.com/unknown-horizons/unknown-horizons/wiki/Translators#using-plain-po-files

I recommend to submit it there. Here you can find some basic instructions https://github.com/unknown-horizons/unknown-horizons/wiki/Translators

PR #2350, opened 2016-01-22
Fixed bug where keyboard scrolling hanged from time to time

Teemperor:
Sometimes scrolling hangs when using the keyboard (especially when navigating fast).

How to reproduce:

1. Press and hold left arrow Key -> View moves to the left which is OK
2. Also press and hold right arrow key -> View still moves to the left which can be tolerated
3. Release left arrow key. -> View stops moving instead of moving right (as right arrow key is still pressed).

This patch resolves this issue.


stubb:
Great work and nice find!

PR #2351, opened 2016-01-23
Generate ai names from preset list

Teemperor:
Numbers as AI names are boring, so I've added a small list of explorers/captains that is used as a template. if we run out of names, we fall back to the old behaviour.

### DIFF
--- a/horizons/util/startgameoptions.py
+++ b/horizons/util/startgameoptions.py
@@ -20,12 +20,27 @@
 # ###################################################
 
 from operator import itemgetter
+from random import shuffle
 
 from horizons.constants import AI, COLORS
 from horizons.util.color import Color
 from horizons.util.difficultysettings import DifficultySettings
 
 class StartGameOptions(object):
+
+	# list of possible ai names. Should somewhen be outsourced to the name database
+	preset_ai_names = [
+		"Christopher Columbus",
+		"William Adams",
+		"Vitus Bering",
+		"John Smith",
+		"Paulo da Gama",
+		"Ferdinand Magellan",
+		"Anna Shchetinina",
+		"John M. Whitall",
+		"Leif Eriksson"
+	]
+
 	def __init__(self, game_identifier):
 		super(StartGameOptions, self).__init__()
 		self.game_identifier = game_identifier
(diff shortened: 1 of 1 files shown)


stubb:
Thanks for your PR, that is a nice addition. But we should move the data into the database. 
Something like:
```SQL

CREATE TABLE "ainames" (
	"name" TEXT NOT NULL,
	"locale" TEXT NOT NULL DEFAULT ('en')
);
```
Within the file content/names.sql. The query should go into the UhDbAccessor class, for example into a method called get_random_ai_name. You can access the global db via horizons.globals.db so you can call it like:
```PYTHON
horizons.globals.db.get_random_ai_name()
```

Teemperor:
Thanks for the suggestion!

I thought about that, but if we want unique names for each AI, we can't just call get_random_ai_name() n times (there is a good chance to get the same name multiple times). We eventually end up doing some unique-index-shuffling-magic or just reading the whole table into a list and then shuffle it as in the current code. Both sounds ugly.

I'll look into it.

LinuxDonald:
@Teemperor any news here?

MasterofJOKers:
I haven't looked at the code, but spontanously had some ideas regarding the mentioned problem.

What about using a temporary table, from which used AI names are deleted? Probably depends on whether all AI names are fetched through the same database connection, if I understand http://www.sqlite.org/tempfiles.html#tempdb correctly.

Another option might be passing `get_random_ai_name()` a list of already used AI names and doing something along the lines of `WHERE name NOT IN $used_list`. This depends on knowing the AIs names, though.

Teemperor:
Sorry, don't have time to finish this, but @MasterofJOKers suggestions sound good.

MasterofJOKers:
@stubb Do we really need a locale for the names? They seem to be universal to me, as they are taken from history.

MasterofJOKers:
@ChrisOelmueller told me on the IRC channel, that the locale was supposed to be optional in the sense, that names, which are only valid in a single locale, should only be displayed there - but names fitting every locale can be used everywhere. Therefore, I'll make the "locale" field nullable.

MasterofJOKers:
Reopened as #2395.

PR #2352, opened 2016-01-23
Fix crash on missing thumbnails in unit queue

Teemperor:
How to reproduce:

1. Build barracks
2. Start building swordsmen
3. Start building another swordsmen
4. Crash due to missing swordsmen thumbnail for the que

This patch fixes this by using a fallback image. I kept the patch generic because we probably get more of these problems when we can produce different units.


LinuxDonald:
We know that it is broken. @stubb is working on the unit production. Thx for the patch. But maybe we will fix the crash directly.

Teemperor:
I'll keep it open for now. Feel free to close it if it superseded by another PR

stubb:
My own solution was providing an thumbnail, but I like the idea of using an exception. Similar approach is used for imagebuttons. @Teemperor I would appreciate a related PR for the fifengine

stubb:
Thanks for this PR :+1:

PR #2348, opened 2016-01-07
Add building barrier. WIP.

stubb:
- [ ] Needs proper description within the object file.
- [ ] Buildmenu icons.
- [ ] GFX
- [ ] Overview Tab should have title, health, description & image.

PR #2353, opened 2016-01-23
Adds a barrier building

stubb:
- [x] Needs proper description within the object file.
- [x] Buildmenu icons.
- [x] GFX
- [x] Overview Tab should have title, health, description & image.
- [x] Use correct images within buildtool
- [x] Use correct images within buildpreview
- [x] Separate PathObject from gfx orientation code
- [ ] Allow wall building on coast

The GFX needs to be 1x1, because there is a "missing feature" within the related code https://github.com/unknown-horizons/unknown-horizons/blob/master/horizons/world/building/buildable.py#L375

Note: Prior PR #2348 closed because of commit fff837ad88314b8cc1c7bb30f34c83e18f879c6c


LinuxDonald:
@stubb we tested it with @janexx grafics but it dont show it correct.

LinuxDonald:
Maybe need clean up in Path code?

squiddy:
@LinuxDonald Do you know what the last two TODOs are about?

>  Separate PathObject from gfx orientation code
 Allow wall building on coast

squiddy:
@LinuxDonald I've taken the liberty to rebase the branch onto master. The code was from february/march, and in the mean time the stone deposit was added which would introduce an ID clash. I've moved the barrier to 71.

The original branch can be found here: https://github.com/unknown-horizons/unknown-horizons/tree/barrier_building_old

MasterofJOKers:
@squiddy Maybe this code is supposed to be separated: https://github.com/unknown-horizons/unknown-horizons/blob/dadeaeef2f463d5bd97eec3178d6d0857fde8b44/horizons/world/building/path.py#L65

squiddy:
Are these barriers supposed to allow movement? The graphics seem to suggest that, e.g. 

If we were to allow it, we will run into issues with ordering of graphics: https://www.youtube.com/watch?v=ShYzu8xzHFU

MasterofJOKers:
I don't think so. Looking at the examples in https://github.com/unknown-horizons/unknown-horizons/blob/dadeaeef2f463d5bd97eec3178d6d0857fde8b44/horizons/world/building/path.py#L65 "abcd" is supposed to be surrounded in all directions by other walls.

squiddy:
Not sure I get that, have you looked at the video? Perhaps it's just a bug, the behaviour of walls next to roads don't look unreasonable though. It essentially would create a gate in a wall where a road goes through.

I'm just wondering if the graphics in this PR that show openings were made on purpose, otherwise we can get rid of these.

MasterofJOKers:
I don't think that's supposed to happen. Definitely a bug to me.

squiddy:
Alright, solid wall it is then. :)

LinuxDonald:
The grafics are maybe by the example of street grafics. Not I never planned that anything can walk on the wall. But the bug here is that the wall don't get build correctly.

squiddy:
> Not I never planned that anything can walk on the wall.

Thanks for the confirmation. Work can continue. :)

> But the bug here is that the wall don't get build correctly.

I'm working on it. The rendering was fixed by subclassing from `Path`, which however led to the strange behaviour with roads. Now it's clear what `Separate PathObject from gfx orientation code` means.

squiddy:
Barriers are properly aligned now. I'm going to add a test somehow and then extract the functionality so both Barrier and Path can make use of it.

squiddy:
@LinuxDonald I appreciate it if you could have a look at the current state of this PR and test it. If it works, I'd suggest to get this into `master` now and deal with building on the coast later.

LinuxDonald:
Okay squidds I will test the pr later or tomorrow. First I must compile fife with python2 support than I can test It :-)

LinuxDonald:
:+1: for merge. Tested it and it works. Needs one more grafic that it can build on streets/over streets. And maybe code for it?

And about: Allow wall building on coast
Should we really make walls buildable on coast?

Issue #2338, opened 2015-12-01
Tooltips width is not correct


Gladius1:
As you may know, the width of the tooltip icon is not correct. For many menu tooltips it is too big, for the building menu it is often too small. I attempted to fix it but I could use some help.
Please have a look at my implementation and give me some hints, so I can make tooltips with correct width.
I'm especially interested if it is possibe to just scale the images in the buffer or if it is possible to scale the icon. This would make things way easier.
https://github.com/Gladius1/unknown-horizons/tree/GuiTooltips

Gladius1:
I made further improvements to the tooltips. It works pretty well now, but pychans textwrapping is bugged. It would cut off words from long strings, so I used the textwrap from python. It is char based. So it wraps after n chars, but since we are NOT using a monospaced font, n chars isn't always n pixels long. So sometimes part of a word gets chopped off. I don't know a fix for this, maybe someone else has an idea.
One solution would be to use a monospaced font for tooltips, but I don't know how to set the font...

https://github.com/Gladius1/unknown-horizons/tree/tooltip-fix-gladius

LinuxDonald:
See https://github.com/unknown-horizons/unknown-horizons/pull/2383

AndyMender:
Will see whether it's possible to split/slice the text and/or adjust the tooltip width based on that. Some single-line tooltips really look strange with the empty space in the widget.

Issue #2227, opened 2014-01-12
Inventory resource hovered message not dismissed


pinkfloyda:
![screenshot from 2014-01-12 22 35 11](https://f.cloud.github.com/assets/5958677/1896336/fef4ee94-7b96-11e3-8ebf-e4d6c2aff808.png)

PR #2365, opened 2016-03-08
Fixed the tooltips

Gladius1:
Finally got the time to fix the tooltips. I hope everything works as intended

### DIFF
--- a/horizons/entities.py
+++ b/horizons/entities.py
@@ -47,7 +47,9 @@ class _EntitiesLazyDict(dict):
 class Entities(object):
 	"""Class that stores all the special classes for buildings, grounds etc.
 	Stores class objects, not instances.
-	Loads everything from the db."""
+	Loads grounds from the db.
+	Loads units and buildings from the object YAML files.
+	"""
 	loaded = False
 
 	log = logging.getLogger('entities')


stubb:
Thanks for this PR :+1: 
Can you elaborate which problems you tried to fix? Maybe there is already an issue covering the topic.

Gladius1:
Well for once the width of the tooltips was off:
 Tooltips width is not correct #2338 

Also the tooltips sometimes wouldn't dissappear. This was fixed for the inventory resource tooltips, but not for the island resource tooltips.
 Inventory resource hovered message not dismissed #2227

Gladius1:
The fix doesn't work well for very small buttons, this will have to be fixed at some time. Maybe when styles are fully implemented in pychan. For now this was the best I was capable of.

Gladius1:
I just realized that it is still possible to get a tooltip stuck by moving the mouse very fast over a button with a tooltip. So this is no complete fix just a mend.

stubb:
Seems like for some tooltips the width does not fit yet.

Gladius1:
I can't replicate this. Are you using the newest commit?

stubb:
Yes, i used your recent commits. Maybe iam using an new version of fifechan/fifengine. Was there a special reason you close this issue?

Gladius1:
I thought that since there are still some fixes to be made, I should commit to a branch and not master. Until all bugs are fixed I'll code on the branch:
https://github.com/Gladius1/unknown-horizons/tree/tooltip-fix-gladius

stubb:
That sounds like a good idea!

Issue #2369, opened 2016-03-11
Production Overview is broken


LinuxDonald:
![2016-03-11 18-59-56 415537](https://cloud.githubusercontent.com/assets/1757651/13710793/860122f8-e7bb-11e5-858e-017d80b284ba.png)

There is nothing showed...

LinuxDonald:
https://github.com/unknown-horizons/unknown-horizons/pull/2372

Jeeppler:
Is this fixed or open?

LinuxDonald:
It's open. I only closed the pull request because that don't become finished.

PR #2370, opened 2016-03-14
Production Overview fix

Gladius1:
I fixed the bug that the production overview was empty.  Production Overview is broken #2369 
Once we have buildings that produce units or ships, this has to be reverted.

### DIFF
--- a/horizons/gui/ingamegui.py
+++ b/horizons/gui/ingamegui.py
@@ -135,9 +135,6 @@ class IngameGui(LivingObject):
 			'logbook' : lambda: self.windows.toggle(self.logbook)
 		})
 		self.mainhud.show()
-		GuiAction.subscribe(self._on_gui_click_action)
-		GuiHover.subscribe(self._on_gui_hover_action)
-		GuiCancelAction.subscribe(self._on_gui_cancel_action)
 
 		hotkey_replacements = {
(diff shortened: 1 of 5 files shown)


stubb:
The boatbuilder and barracks buildings are unit/ship producer. Your fix simply tries to display all produced resources including units now. I think we need a better approach to fix this. Your help is really appreciated!

Gladius1:
This should work. The bug was, that all components had the name "producer". Since the names were equal the condition wasn't met and the def to update the settlement inventory wasn't called.

Gladius1:
Please check out my tooltip fix as well.

https://github.com/Gladius1/unknown-horizons/tree/tooltip-fix-gladius

stubb:
@Gladius1 please open another PR for the tooltip fix

stubb:
About the overview. It does not work, because it tries to display all produces "resouces" now. Including the units. A good solution would be to only display produced resources from Producer not UnitProducers of any kind.

stubb:
Sorry, did not saw your last commit. But this introduces other issued where we check if a building has the component Producer, like here https://github.com/unknown-horizons/unknown-horizons/blob/master/horizons/world/building/buildingresourcehandler.py#L53. Components are identified by the name. Your commit introduces two new components and the check fails for them. This needs a better solution then simply looking for each component if it is available or not.

PR #2383, opened 2016-05-25
Tooltip fix gladius [NEED TO BE FIXED]

LinuxDonald:
@Gladius1 are you done on this?

### DIFF
--- a/horizons/gui/widgets/tooltip.py
+++ b/horizons/gui/widgets/tooltip.py
@@ -20,26 +20,28 @@
 # ###################################################
 
 import re
-
+import textwrap
 
 from fife import fife
 from fife.extensions.pychan.widgets import HBox, Icon, Label
 
 import horizons.globals
 
-from horizons.extscheduler import ExtScheduler
 from horizons.gui.util import get_res_icon_path
 from horizons.gui.widgets.container import AutoResizeContainer
 from horizons.gui.widgets.icongroup import TooltipBG
+from horizons.messaging import HoverInstancesChanged
 
 class _Tooltip(object):
 	"""Base class for pychan widgets overloaded with tooltip functionality"""
 
-	LABEL_LEFT_RIGHT_MARGIN = 10
+	LABEL_LEFT_MARGIN = 7
+	LABEL_RIGHT_MARGIN = 7
 	# Left and right margin for the tooltips text, otherwise the text is over the border of the icon
 	LABEL_TOP_BOTTOM_MARGIN = 5
 	# Top and bottom margin for the tooltips text, otherwise the text is over the border of the icon
 	MAX_LABEL_WIDTH = 145  # Width of the tooltip text before the text wraps
+	MAX_BG_WIDTH = 0
 	# Find and replace horribly complicated elements that allow simple icons.
 	icon_regexp = re.compile(r'\[\[Buildmenu((?: \d+\:\d+)+)\]\]')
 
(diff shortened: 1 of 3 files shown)


LinuxDonald:
@stubb must @Gladius1 fix here something?

horizons/gui/widgets/tooltip.py
```diff
@@ -97,103 +101,98 @@ def position_tooltip(self, event):
...
-		def get_top(w):
-			return get_top(w.parent) if w.parent else w
-		top_pos = get_top(self).position
-		if top_pos == (0, 0):
-			return
+		# def get_top(w):
+		# 	return get_top(w.parent) if w.parent else w
+		# top_pos = get_top(self).position
+		# if top_pos == (0, 0):
+		# 	return
```
MasterofJOKers: If this code is no longer in use, why keep it (commented out)?

horizons/gui/widgets/tooltip.py
```diff
@@ -97,103 +101,98 @@ def position_tooltip(self, event):
...
+
 		if not self.tooltip_shown:
 			self.show_tooltip()
-			#ExtScheduler().add_new_object(self.show_tooltip, self, run_in=0.3, loops=0)
-			self.tooltip_shown = True
 
 	def show_tooltip(self):
-		if not self.helptext:
-			return
+
```
MasterofJOKers: unnecessary empty line?

horizons/gui/widgets/tooltip.py
```diff
@@ -97,103 +101,98 @@ def position_tooltip(self, event):
...
-				hbox.addChildren(icon, label)
+				hbox.addChildren(icon, resource_amount_label)
 			hbox.adaptLayout()
-			# Now display the 16x16px "required resources" icons in the last line.
+			if hbox.width >= self.MAX_LABEL_WIDTH:
+				#self.MAX_LABEL_WIDTH = hbox.width
+				self.wrap_tt_text(buildmenu_icons, tooltip_text)
+				self.MAX_BG_WIDTH = hbox.width
+			else:
+				self.wrap_tt_text(buildmenu_icons, tooltip_text)
```
MasterofJOKers: If the same thing is done in `if` and in `else`, why not just doing it like that:
```python
self.wrap_tt_text(buildmenu_icons, tooltip_text)
if hbox.width >= self.MAX_LABEL_WIDTH:
  self.MAX_BG_WIDTH = hbox.width
```

horizons/gui/widgets/tooltip.py
```diff
@@ -97,103 +101,98 @@ def position_tooltip(self, event):
...
+
+	def wrap_tt_text(self, buildmenu_icons, text):
+		chars_per_line = 1
+		for i in range(0, text.__len__()):
+
+			if self.label.text.__sizeof__() < self.MAX_LABEL_WIDTH:
+				self.label.text += text[i]
+				chars_per_line += 1
+			self.gui.adaptLayout()
+		chars_per_line - 1
```
MasterofJOKers: This line does nothing at all O.o Maybe there's an `=` missing? But then how could it be needed, if everything worked like this?

horizons/gui/widgets/tooltip.py
MasterofJOKers: Why is `self.MAX_BG_WIDTH` set here explicitly? Can you please explain this to me? It is also set in `self.wrap_tt_text()`. That's why I'm wondering. Maybe you could add a comment, why?

horizons/gui/widgets/tooltip.py
MasterofJOKers: Just for my understanding: There are chars set, to see how wide the label will get, because we don't know the font information, so we cannot calculate it in another way. Right?
The later, we will use the found information to shorten the text to the number of chars. But since this is probably not monospace, different lines might have widely differing lengths because of different contained characters. Therefore, the text could still overlap, if I see this correctly.

LinuxDonald:
Closed for now see Ticket https://github.com/unknown-horizons/unknown-horizons/issues/2338

PR #2395, opened 2016-06-25
Better AI names

MasterofJOKers:
Citing @Teemperor in its PR #2351:
> Numbers as AI names are boring, so I've added a small list of explorers/captains that is used as a
> template. if we run out of names, we fall back to the old behaviour.

This includes the changes to PR #2351, that @stubb expected to be made.


MasterofJOKers:
I don't understand, why the tests fail. Doesn't seem to be cause by this commit(s). It works locally:
```
$ python run_tests.py --with-xunit --verbose --nologcapture
....
Ran 157 tests in 44.877s

OK (SKIP=2)
```

LinuxDonald:
What about merging this?

MasterofJOKers:
Don't know, should I do this?

LinuxDonald:
Is it complete? Than yes please.

PR #2398, opened 2016-06-27
Use contextmanager when opening a file

MasterofJOKers:
Using the with-statement to open a file ensures it is closed even on
raised exceptions. Additionally, one cannot forget to call `.close()`.


LinuxDonald:
I´m new in python but it looks okay for me +1 for merge.

Issue #2388, opened 2016-05-29
Settler stages grafic´s dont show up ingame


LinuxDonald:
The settler stages grafic´s dont show up ingame. Only its shows the sailer grafics.

TIER.PIONEERS: and TIER.SETTLERS: dont show up.

stubb maybe you can find an fix that Jana can render all martins grafics and add them into the game :)

https://github.com/unknown-horizons/unknown-horizons/blob/master/content/objects/units/collectors/settlercollector.yaml

LinuxDonald:
Janexx have tested it when you comment out TIER.PIONEERS: and TIER.SETTLERS: it loads the grafics as Sailors one but the are in other folders see: https://github.com/unknown-horizons/unknown-horizons/tree/master/content/gfx/units/inhabitant

I think it dont check for Tier and that is maybe the problem. That it is not implemented.

MasterofJOKers:
@LinuxDonald You're right, it is not implemented for units to be updated with the tier. Should I implement this just for the inhabitants or for other units as well?

LinuxDonald:
i think there must be an way that the game checks in the yaml file. If there are defined grafics for the tier when yes that it load it. But the game must check what tier the buildings have too and that one building get an higher tier and than only runs on streets the newer tier grafics of people.

MasterofJOKers:
Well, I can tell you, how it is done for buildings: When a new level is reached, every building is informed about it and tryies to load a new random action_set into itself. This can also be done for the buildings' collectors, which inhabitants are - but since tents behave a little different in that they initiate the level change, there are two places to implement having different grafics for collectors of buildings.

LinuxDonald:
Fixed with https://github.com/unknown-horizons/unknown-horizons/commit/510fe098a7d12fdfb2fa4dd3ca0457e683d649d4

PR #2399, opened 2016-06-30
Allow different collector unit graphics per tier

MasterofJOKers:
When a new tier is reached, a message is broadcasted to all buildings,
which lets them upgrade their level and with that their graphics.
Collectors are units, that belong to a building. This commit lets a
building tell its collectors, that its level has changed and their level
should be changed, too. Other than that, `BuildingCollector`s now also
have a `level_upgrade()` method to update their current `_action_set_id`
to a random one of the new level.

The corresponding issue is #2388.

### DIFF
--- a/horizons/world/building/building.py
+++ b/horizons/world/building/building.py
@@ -24,6 +24,7 @@ import logging
 from fife import fife
 
 from horizons.command.building import Build
+from horizons.component.collectingcomponent import CollectingComponent
 from horizons.component.storagecomponent import StorageComponent
 from horizons.component.componentholder import ComponentHolder
 from horizons.constants import RES, LAYERS, GAME
@@ -203,6 +204,12 @@ class BasicBuilding(ComponentHolder, ConcreteObject):
 		"""Upgrades building to another tier"""
 		self.level = lvl
 		self.update_action_set_level(lvl)
+
+		# any collectors (units) should also be upgraded, so that their
+		# graphics or properties can change
+		if self.has_component(CollectingComponent):
+			for collector in self.get_component(CollectingComponent).get_local_collectors():
+				collector.level_upgrade(lvl)
 
 	@classmethod
 	def get_initial_level(cls, player):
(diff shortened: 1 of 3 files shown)


LinuxDonald:
https://gist.github.com/LinuxDonald/f67de466666c33e3b4684616a972bf00

I think the i need some fixing or something is wrong with the pr?

i had now two times this crash.

MasterofJOKers:
Thank you for trying it. I'll look into this on the weekend.

MasterofJOKers:
@LinuxDonald I've tried it. The AI successfully updated one tier and I had no crash. It didn't look like it would be my code changes in the backtrace in the first place, but I wanted to be sure. Can you please try this savegame and see if it crashes? I've run it on speed 11x or something and it didn't crash for me. [savegame](http://joker.someserver.de/uh/ai_tier.sqlite.gz)

MasterofJOKers:
Ok, got it after some time in another game with 3 AI players. What was your setup for getting the crash?

LinuxDonald:
I've got the crash with 3 and 4 ai players. Not tested it with 2 or 1 ai player.

MasterofJOKers:
Do you still have a savegame of this?

MasterofJOKers:
So I've got a savegame now, right before it crashes. It crashes with and without my patch. But the savegame is created with my patch, so there might be internal data loaded, that causes it. But I can now probably debug it.

LinuxDonald:
That's listen good :-)

LinuxDonald:
Now i have become the crash when the second ai player want to build an "Kontor". On map: quattro. With 3 Ai players. That crash is really not normal it hart to reconstruct it...

MasterofJOKers:
OK, then I have to try that. You don't get the crash, if you are running on master, right?

MasterofJOKers:
I've started a game twice using quattro with 3 AIs - nothing happened. Neither on master nor on this branch.

LinuxDonald:
On master I don't have any problems. Maybe the new code must be synced with ai code?

LinuxDonald:
Maybe this should be tested with latest master code.

MasterofJOKers:
You mean to see, if it /still/ works? There have been no changes that I'm aware of, that were in the same place, so it should still just work.

LinuxDonald:
+1 for merge

PR #2400, opened 2016-07-01
remove double import AbstractClayDeposit

MarkusHackspacher:
Tow times import AbstractClayDeposit, in line 57 and 75.


LinuxDonald:
Thx :-)

PR #2402, opened 2016-07-02
use `.format()` function and staticmethod

MarkusHackspacher:


LinuxDonald:

MasterofJOKers:
The same comment as on the other pull-request also applies hier:
* don't make changes you're not telling about
* use up to 90 characters

Please integrate the `.format()`-parts of this commit into the other pull request. It would be also better, to make an own commit for the `staticmethod` change, which seems appropriate, as the class parameter is not used. But if such a change is done, than my view is, that one should look through the whole project for `classmethod`s, that can be changed into `staticmethod`s instead of in only one file. Making the change in the whole project creates a consistent code-base, while changing only one file does not.

MarkusHackspacher:
The format part is now in the other commit.

And the classmethod vs. staticmethod: I would not make the whole project inconsistent.
It is better to close this comment.

MasterofJOKers:
I'm sorry to hear that. I would have liked to see the changes made :(
But thank you for moving the format part into the other pull-request.

MarkusHackspacher:
@MasterofJOKers:
line 95 and 96:
Orginal:
-	@classmethod
-	def load_yaml_data(self, string_or_stream):
my solution:
+	@staticmethod
+	def load_yaml_data(string_or_stream):

maybe a other solution(cls for self):
-	@classmethod
-	def load_yaml_data(cls, string_or_stream):

The other classmethod s in the file use always cls for self.

MasterofJOKers:
Right, it should not be `self`. Doing a little grep shows other places, that don't use `self`. Would you like to clean that up e.g. make that more consistent?
```
$ git grep -A 1 @classmethod | grep -B 1 self
ai/aiplayer/building/farm.py:	@classmethod
ai/aiplayer/building/farm.py-	def _suitable_for_road(self, production_builder, coords):
--
command/__init__.py:	@classmethod
command/__init__.py-	def allow_network(self, klass):
--
manager.py:	@classmethod
manager.py-	def allow_network(self, klass):
--
network/packets/__init__.py:	@classmethod
network/packets/__init__.py-	def add(self, origin, klass):
--
network/packets/__init__.py:	@classmethod
network/packets/__init__.py-	def set_mode(self, client=True):
--
network/packets/__init__.py:	@classmethod
network/packets/__init__.py-	def find_class(self, module, name):
--
network/packets/__init__.py:	@classmethod
network/packets/__init__.py-	def loads(self, str):
--
util/worldobject.py:	@classmethod
util/worldobject.py-	def get_objs(self): return self.__objects
--
util/yamlcache.py:	@classmethod
util/yamlcache.py-	def load_yaml_data(self, string_or_stream):
```

MarkusHackspacher:
Yes, I can do it. 9 times in 6 files, like Point 8.
https://www.quantifiedcode.com/app/project/dfb8434844b0481d9328975c88c8d703

MasterofJOKers:
Oh man, I don't even want to look at that link. Not everythin needs fixing, though ... But the first entry? A for-loop with an `else` but without a `break`? That looks just wrong O.o

PR #2403, opened 2016-07-03
classmethod should be called `cls` and not `self`

MarkusHackspacher:

I make for each file a commit with the same name.

### DIFF
--- a/horizons/ai/aiplayer/building/farm.py
+++ b/horizons/ai/aiplayer/building/farm.py
@@ -263,8 +263,12 @@ class FarmEvaluator(BuildingEvaluator):
 		cls.__field_offsets = first_class + second_class + third_class
 
 	@classmethod
-	def _suitable_for_road(self, production_builder, coords):
-		return coords in production_builder.land_manager.roads or (coords in production_builder.plan and production_builder.plan[coords][0] == BUILDING_PURPOSE.NONE)
+	def _suitable_for_road(cls, production_builder, coords):
+		"""check coordinates
+		returns true/false"""
+		return coords in production_builder.land_manager.roads or (
+			coords in production_builder.plan and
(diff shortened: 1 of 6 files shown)


horizons/ai/aiplayer/building/farm.py
```diff
@@ -263,8 +263,12 @@ def init_field_offsets(cls):
 		cls.__field_offsets = first_class + second_class + third_class
 
 	@classmethod
-	def _suitable_for_road(self, production_builder, coords):
-		return coords in production_builder.land_manager.roads or (coords in production_builder.plan and production_builder.plan[coords][0] == BUILDING_PURPOSE.NONE)
+	def _suitable_for_road(cls, production_builder, coords):
+		"""check coordinates
+		returns true/false"""
```
MasterofJOKers: If that comment is really necessary (in my opinion, its not), then it should bue `True/False`.

horizons/network/packets/__init__.py
```diff
@@ -60,11 +60,13 @@ class SafeUnpickler(object):
 	- http://nadiana.com/python-pickle-insecure
 	"""
 	@classmethod
-	def add(self, origin, klass):
+	def add(cls, origin, klass):
+		"""Adding SafeUnpickler to the pickle whitelist
+		"""
```
MasterofJOKers: Please put the closing `"""` on the end of the previous line.

horizons/util/worldobject.py
```diff
@@ -82,4 +82,6 @@ def __lt__(self, other):
 
 	# for testing:
 	@classmethod
-	def get_objs(self): return self.__objects
+	def get_objs(cls):
+		"""returns objects"""
```
MasterofJOKers: This comment is not necessary, as it does not add any explanation, but just repeats the method name.

MasterofJOKers:
Thank you very much.

PR #2412, opened 2016-07-23
change to the `.format()` function Part 2

MarkusHackspacher:
All % are now changed. I used a change function at https://www.quantifiedcode.com


LinuxDonald:
@MasterofJOKers could you maybe review this please?

MasterofJOKers:
Is this done? Last change was 3 days ago ...

LinuxDonald:
@MarkusHackspacher are you done with this?

MarkusHackspacher:
I test with test `bash development/create_pot.sh` and found some bugs, but don´t know sould be the line ´#. (itstool) path: VBox/Label@text´ remove or not, was it a fault before or not.

po/uh/unknown-horizons.pot
```diff
@@ -1,222 +1,232 @@
...
 msgid "Return to editor"
 msgstr ""
 
-#. (itstool) path: VBox/Label@text
 #: content/gui/xml/editor/editor_pause_menu.xml:15
+#: horizons/gui/modules/select_savegame.py:53
 msgid "Save map"
 msgstr ""
 
-#. (itstool) path: VBox/Label@text
```
MarkusHackspacher: If it ok to remove or not?

MasterofJOKers: Where does it come from? Do you think it serves a purpose?

MasterofJOKers:
Afaics, the comment shows, what object the text is taken from in the xml file. So I guess it should not be removed.

development/extract_strings_from_objects.py
```diff
@@ -186,8 +185,8 @@ def add_line(value, component, sep, key, filename):
 	strings = sorted(object_strings)
 
 	if strings:
-		return(u'\n\t"{0!s}" : {{{1!s}{2!s},{3!s}}'.
```
MasterofJOKers: So you remove the unicode marker here, but add it in the above diff and remove it in the one above that one. Can you explain, why?

MarkusHackspacher:
I change this line back to the old % operator, I make a fault in this case. But I see now, there a '{' too much.
I´ll look, and send a commit.

po/uh/unknown-horizons.pot
MarkusHackspacher: It is OK, it is the same like the actual code. Maybe too much changes in last time. I got a marge error at last.

https://github.com/unknown-horizons/unknown-horizons/blob/master/po/uh/unknown-horizons.pot

MasterofJOKers:
Now I get it.
```python
>>> ('\n\t"%s" : {' % filename) + (ROWINDENT + '%s,' % (','+ROWINDENT).join(strings)) + ROWINDENT + '},'
'\n\t"asdf.bla" : { bla, blubb, },'
```
is the same as
```python
>>> '\n\t"{filename}" : {{{rowindent}{strings},{rowindent}}},'.format(filename=filename, rowindent=ROWINDENT, strings=(','+ROWINDENT).join(strings))
'\n\t"asdf.bla" : { bla, blubb, },'
```

So basically, you're just missing a `}` at the end to escape the closing curly brace.

MarkusHackspacher:
I change the line, thank you.
Maybe, have you a solution for 
`object_strings.append('# %s' %comment + ROWINDENT + '%-30s: %s' % (('"%s"') % component, text))`

My try:
`object_strings.append(u'# {0!s} {1!s}"{2:<30!s}": {3!s}'. format(comment,ROWINDENT, component, text))`

https://github.com/MarkusHackspacher/unknown-horizons/blob/master/development/extract_strings_from_objects.py#L172

MasterofJOKers:
I'll look at it tomorrow. Don't want to think any more today.

MarkusHackspacher:
`%-30s` means every time 30 long right will fill with space. 
I I remove -30s and the result is better for reading in the file 
horizons/i18n/objecttranslations.py.
https://github.com/unknown-horizons/unknown-horizons/pull/2412/files#diff-3e32c78b350b10afbd13f7be2daabf0b

If it ok, you can marge, I´m finish.

MasterofJOKers:
You should not change the output when changing to `format()`. If the output should be changed, open a new pull request for or **at least** do it in a different commit.

`%-30s` can be replaced by `{0!s:<30}`.

MarkusHackspacher:
@MasterofJOKers could you check?

MasterofJOKers:
Status: I currently don't have the time/motivation for such a large merge. The last merge took me > 1h and that was a much smaller diff.

MarkusHackspacher:
Could it help, if I split up this big PR in smaller parts?
I will https://patch-diff.githubusercontent.com/raw/unknown-horizons/unknown-horizons/pull/2412.diff
copy ervery 10 changes in a file (format2-1.patch, format2-2.patch ..) and make a separate PR per file with `git apply format2-1.patch`

MasterofJOKers:
No. I just need to find the time. It will be more consistent that way. I'm sorry, that it takes so long.

MasterofJOKers:
Working on it. Got through development and `horizons.ai` today.

development/extract_strings_from_objects.py
```diff
@@ -161,12 +161,13 @@ def add_line(value, component, sep, key, filename):
 			text = u'_("{value}")'.format(value=value[2:])
 			component = component + sep + str(parse_token(key, 'TIER'))
 			filename = filename.rsplit('.yaml')[0].split(OBJECT_PATH)[1].replace('/',':')
-			comment = '%s of %s' %(component, filename)
-			object_strings.append('# %s' %comment + ROWINDENT + '%-30s: %s' % (('"%s"') % component, text))
+			comment = u'{0!s} of {1!s}'.format(component, filename)
+			object_strings.append('# {comment!s}{rowindent}{component:<30}: {text!s}'.
+				format(comment=comment, rowindent=ROWINDENT, component='"{}"'.format(component), text=text))
```
squiddy: I'd drop the `component='"{}"'.format(component)` (replace it with `component=component` and instead include the quotes in the original string.

`'# {comment!s}{rowindent}"{component:<30}": {text!s}'`

MasterofJOKers:
You'd need it to explicitly tell, that you don't want `!r` ;) Other than that, I was beginning to ask myself the same question.

squiddy:
> You'd need it to explicitly tell, that you don't want !r ;) Other than that, I was beginning to ask myself the same question.

Not sure I can follow. To me it looks like !s is the default.

```python
>>> print '{}'.format("foobar")
foobar
>>> print '{!s}'.format("foobar")
foobar
>>> print {!r}'.format("foobar")
'foobar'
```

MasterofJOKers:
It is the default. That's why I said, you need it to *explicitly* tell, what you want.

MasterofJOKers:
Also, we can't just use `{}` with python 2.6 :(

LinuxDonald:
Don't have debain python 2.7?

siccegge:
In oldstable already ;-)

LinuxDonald:
Then set min python version 2.7 and don't support 2.6.

po/uh-server/unknown-horizons-server.pot
```diff
@@ -1,14 +1,14 @@
 # SOME DESCRIPTIVE TITLE.
 # Copyright (C) YEAR The Unknown Horizons Team
-# This file is distributed under the same license as the Unknown Horizons Server package.
+# This file is distributed under the same license as the PACKAGE package.
```
MarkusHackspacher: don´t merge, I´ll for this: for PACKAGE should be Unknown Horizons Server.

development/extract_strings_from_objects.py
MarkusHackspacher: That we have discus: 

https://github.com/unknown-horizons/unknown-horizons/pull/2412#issuecomment-240862806
%-30s means every time 30 long right will fill with space.
I I remove -30s and the result is better for reading in the file
horizons/i18n/objecttranslations.py.

You should not change the output when changing to format(). If the output should be changed, open a new pull request for or at least do it in a different commit.

po/uh-server/unknown-horizons-server.pot
MarkusHackspacher: I do not see something where the output change. Maybe my xgettext version (0.19.3) is too old..

squiddy:
@MarkusHackspacher  What's the status here? Is this done?

MasterofJOKers:
@squiddy Yes. It is my responsibility to merge this. I've looked through the AI-stuff already, as I said.

squiddy:
It says "DONT MERGE" here: https://github.com/unknown-horizons/unknown-horizons/pull/2412#discussion_r79554277. And I was wondering if there was an update or if we could help.

MasterofJOKers:
After some thought, I came to the conclusion, that @squiddy is right: we don't need the `!s` in every string.

Also, we don't gain much from converting `%s` to `{0}`. It would help translators much more, if we used names like `'{asdf}'.format(asdf='bla')`. @MarkusHackspacher Would you be willing to make these changes? Sorry for making up my mind so late.

MarkusHackspacher:
At the moment I do not have so much time, because I am looking for another job. I would be able to automatically search and replace the !s, but then the part of 1 is also replaced and the codebase is already old and probably brings only more conflicts. And more work for you to solve this.

My suggestion take change over in the current code basis and than search and replace !s.

MasterofJOKers:
Search and replace !s would not be enough, as nearly all strings have to be touched by hand, I guess, to add variable names that make sense.

Since we now know what the result should look like, smaller batches of changes would be appreciated. I probably don't have the time for reviewing/merging so much in a single run.

MasterofJOKers:
Oh, and to make it clear: We really don't need to change debug output, that isn't translated to the new format.

MarkusHackspacher:
> Since we now know what the result should look like, smaller batches of changes would be appreciated. I probably don't have the time for reviewing/merging so much in a single run.

I think so, too. Let´s close this PR and start new from the actual code base.

PR #2415, opened 2016-08-03
Some changes to: add_weblate_language.sh

LinuxDonald:
Fixed add_weblate_language.sh that it works from development folder
Added command for create The_Unknown scenario po files

@MasterofJOKers please review it and when you want expand it

### DIFF
--- a/development/add_weblate_language.sh
+++ b/development/add_weblate_language.sh
@@ -5,19 +5,25 @@ if [[ "$#" -ne 1 ]]; then
 	exit 1
 fi
 
-f="po/uh/unknown-horizons.pot"
-o="po/uh/${1}.po"
+f="../po/uh/unknown-horizons.pot"
+o="../po/uh/${1}.po"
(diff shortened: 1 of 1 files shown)


MasterofJOKers:
Two things:
* Why? You could just run `development/add_weblate_language.sh` instead. Keeping it that way would also support moving the script to another place like e.g. `development/translation/`. I think the assumption, that a script is running from the root-folder of UH is not a bad one.
* You have two distinct/separate changes in the same commit. It should be two commits. One for the first line of you commit message and the second one for the second line of you commit message.

MasterofJOKers:
Other than that: We should not use explicit names for the scenarios, but just copy all scenarios. So please convert this into a for-loop. This would also get rid of the second, not needed `mkdir -p`.

MasterofJOKers:
Sorry, just read, that I'm supposed to expand it. Will do it later today, I guess.

LinuxDonald:
Thx @MasterofJOkers :-)

PR #2422, opened 2016-08-12
Avoid using `len(x)` to check if a list is empty

MarkusHackspacher:
I change the code to make the running time faster.
Travis need for the code check now 49.242s to 49.900s. The other check need 61s to 69s.
I hope that it was not an (travis)exception, it can anyone confirm?

### DIFF
--- a/horizons/engine/engine.py
+++ b/horizons/engine/engine.py
@@ -233,7 +233,7 @@ class Fife(object):
 		if remkey in old_keys:
 				old_keys.remove(remkey)
 		if not old_keys:
-				print 'Cannot have no binding for action'
+				print('Cannot have no binding for action')
 				return
 		self.set_key_for_action(action, old_keys)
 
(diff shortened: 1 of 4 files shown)


LinuxDonald:
When travis I'd faster I'd than the game maybe a bit faster too? MasterofJOKers is now 2 weeks not here (vacation)

LinuxDonald:
I found the len is an python fuction https://docs.python.org/2/library/functions.html#len is len here needed or can it be really removed?

MarkusHackspacher:
https://www.quantifiedcode.com/app/issue_class/3LKECaRv
or https://doc.odoo.com/contribute/15_guidelines/coding_guidelines_python/#collections-are-booleans-too

LinuxDonald:
thx for the info i will test the change later. If there are no problems i will merge that :)

Thx for your work :+1:

LinuxDonald:
Found no problems :) Thank you for the patch :+1:

PR #2431, opened 2016-08-15
Changed some print to print()

jorgesumle:


LinuxDonald:
This change must be reviewed by @MasterofJOKers but he is in vacation. For aroud 2 weeks.

MasterofJOKers:
We don't do that anywhere, except in one development script. So to keep the code consistent, I would like to reject this. Also, while this seems to work in python2.7 for me, I am not sure whether this would work without `from __future__ import print_function` in python2.6.

PR #2435, opened 2016-08-17
Improved the english text in the tutorial

jorgesumle:
Note I'm changing lower case names to upper case only when they refer to an element of the game to avoid confusion.


LinuxDonald:
Thx for this :+1:

Issue #2434, opened 2016-08-16
Button to restart game after adjusting Settings when restart required? 


jorgesumle:
What about asking the user if he wants to restart the game, when changes need a restart? We could leave the tick icon which is already there, and add a close button.
![captura de pantalla de 2016-08-16 13 31 31](https://cloud.githubusercontent.com/assets/14212780/17697632/2c60b764-63b6-11e6-9400-30d7fe2df03e.png)

jorgesumle:
I got it working: https://github.com/jorgesumle/unknown-horizons. However, when you mess with changing to fullscreen and restarting the game from my prompt the os resolution is messed up. Am I'm missing something? How should I close Unknown Horizons properly?
![captura de pantalla de 2016-08-16 17 34 33](https://cloud.githubusercontent.com/assets/14212780/17705253/07bcfe96-63d8-11e6-9aa1-8f440153ab5e.png)
![captura de pantalla de 2016-08-16 17 34 55](https://cloud.githubusercontent.com/assets/14212780/17705271/1113d4e2-63d8-11e6-8f06-3c9fb6d99a4c.png)

LinuxDonald:
Maybe say UH that is should close and excute run_uh.py ?
Or look how run_uh.py starts uh.

jorgesumle:
I only have to exit fullscreen mode before restarting the game, but how do I do that?

LinuxDonald:
Maybe that can help you? https://github.com/fifengine/fifengine/search?utf8=%E2%9C%93&q=setFullScreen

@stubb could you maybe help here please?

LinuxDonald:
https://github.com/fifengine/fifengine/issues/521
https://github.com/fifengine/fifengine/issues/331

maybe this helps?

jorgesumle:
Cool, it worked. If the pull request is accepted, you can close this issue.

PR #2436, opened 2016-08-17
Made issue #2434 improvement

jorgesumle:
issue URL: https://github.com/unknown-horizons/unknown-horizons/issues/2434

### DIFF
--- a/horizons/gui/modules/settings.py
+++ b/horizons/gui/modules/settings.py
@@ -181,7 +181,8 @@ class SettingsDialog(PickBeltWidget, Window):
 
 		if restart_required:
 			if self.restart_promt():
-				os.execl(sys.executable, sys.executable, *sys.argv)
+				horizons.globals.fife.engine.destroy()
+				os.execv(sys.executable, [sys.executable] + sys.argv)
 
 		self._windows.close()
 
--- a/horizons/gui/modules/settings.py
+++ b/horizons/gui/modules/settings.py
@@ -20,6 +20,8 @@
 # ###################################################
 
 import collections
+import os
+import sys
 
 from fife import fife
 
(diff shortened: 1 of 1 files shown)


LinuxDonald:
I will test that tomorrow :-)

LinuxDonald:
@MasterofJOKers could you maybe do an code review here? I will test it tomorrow.

horizons/gui/modules/settings.py
```diff
@@ -172,12 +175,15 @@ def apply_settings(self):
...
-			self.show_restart_popup()
-
 		self.hotkey_interface.save_settings()
 		self._settings.apply()
 		self._settings.save()
+
+		if restart_required:
+			if self.restart_promt():
+				horizons.globals.fife.engine.destroy()
+				os.execl(sys.executable, sys.executable, *sys.argv)
```
MasterofJOKers: From reading the [documentation](https://docs.python.org/2/library/os.html#os.execl), `os.execv` should be used here.

MasterofJOKers:
Other than that, it looks like it should work.

jorgesumle:
os.execv works fine here. You mean with `os.execv(sys.executable, sys.argv)`?

MasterofJOKers:
No, I mean
```python
args = [sys.executable] + sys.argv
os.execv(sys.executable, args)
```
The following part in the documentation looked like it shouldn't be used that way:
> The “l” variants are perhaps the easiest to work with if the number of parameters is fixed when the code is written;

Why would they write something like that with Python supporting `*args`? Maybe I'm interpreting too much into that sentence, but looking at the documentation, `os.execl()` seems to be the wrong choice - but since it seems to work, I'm not totally against using it. It just feels wrong to me.

jorgesumle:
Yes, that should work too. In the end, os.execl calls os.execv so they are practically the same.

MasterofJOKers:
Please don't edit your comments in a totally meaning-changing way.

PR #2440, opened 2016-08-23
Fix translate

LinuxDonald:
This should fix that some strings was not translatable.

@jorgesumle i hope this will fix the problem :)

### DIFF
--- a/development/update_translations.py
+++ b/development/update_translations.py
@@ -112,7 +112,7 @@ def update_authors_per_file(input_po, regexp=LANG_RE, since='weblate-credits..',
 def main():
 	# Main interface translation (old 'uh' project in pootle)
 	for f in INTERFACE_TRANSLATIONS:
-		#update_from_template(f, INTERFACE_TEMPLATE)
+		update_from_template(f, INTERFACE_TEMPLATE)
 		update_authors_per_file(f)
 
 	# MP server message translation (old 'mp-server' project in pootle)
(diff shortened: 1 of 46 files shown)


LinuxDonald:
@ChrisOelmueller why you comment line 71 here? 8cc6aeab4c25a3305b9e7b274a9934610b1fbada

I uncomment it and now the scripts find more strings.

ChrisOelmueller:
The way I remember it, I wanted more control over when/whether the
commented commands in the script were executed, so I'd uncomment them
locally as needed.

LinuxDonald:
Thx for the information :-) I will merge that later today.

MasterofJOKers:
We could add commandline options for the commented commands, so control is there and no commenting has to happen.

LinuxDonald:
@MasterofJOKers feel free to do an improvemd here. For now i merged only the updatet translations.

jorgesumle:
But if you add the # back, you have to remove it again when you want to update the interface translations, don't you? If you let it uncommented it should works always.

LinuxDonald:
This line is only needed if we add for example new building to the game. I want to wait what @MasterofJOKers want todo here.

squiddy:
What's the status here? @LinuxDonald you mentioned that the translations are already merged, is this only about the commented/uncommented line? Anyone working on this or can it be closed?

MasterofJOKers:
Looking at it, I'm supposed to do something: add commandline options to update_translations.py. I'll open a ticket for that.

PR #2457, opened 2016-08-30
Random fish distribution with a fixed ratio of fish deposits to size …

LinuxDonald:
…of island coastline.

MAybe test this and than merge it?

### DIFF
--- a/horizons/world/worldutils.py
+++ b/horizons/world/worldutils.py
@@ -265,15 +265,16 @@ def add_nature_objects(world, natural_resource_multiplier):
 				coastline_tiles.append(tile)
 		
 		#island_fish_ratio is the max number of fish deposits placed by an island, len(sorted(island.ground_map.iteritems()))/50 means 1/50 coastline tiles recieve a fish deposit
-		island_fish_ratio = len(sorted(island.ground_map.iteritems()))/50
+		island_fish_ratio = len(island.ground_map) / 50
 		
 		for i in xrange(0, island_fish_ratio):
 			tile = coastline_tiles[i*(len(coastline_tiles)/island_fish_ratio)]
 			fish_deposits = 0
 			while fish_deposits == 0:
 				#randomly choose 2 directions to place the fish deposit away from the coastline
-				x_dir = fish_directions[world.session.random.randint(0, 8)][0] + fish_directions[world.session.random.randint(0, 8)][0]
-				y_dir = fish_directions[world.session.random.randint(0, 8)][1] + fish_directions[world.session.random.randint(0, 8)][1]
+				randint = world.session.random.randint
+				x_dir = fish_directions[randint(0, 8)][0] + fish_directions[randint(0, 8)][0]
+				y_dir = fish_directions[randint(0, 8)][1] + fish_directions[randint(0, 8)][1]
 				
 				fish_x = tile.x + x_dir * world.session.random.randint(3, 9)
 				fish_y = tile.y + y_dir * world.session.random.randint(3, 9)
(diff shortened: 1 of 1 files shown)


horizons/world/worldutils.py
```diff
@@ -258,18 +259,28 @@ def add_nature_objects(world, natural_resource_multiplier):
...
-						Build(FishDeposit, fish_x, fish_y, world,
-						      45 + world.session.random.randint(0, 3) * 90,
-						      ownerless=True)(issuer=None)
+		
+			#list all coastline tiles
+			if 'coastline' in tile.classes:
+				coastline_tiles.append(tile)
+		
+		#island_fish_ratio is the max number of fish deposits placed by an island, len(sorted(island.ground_map.iteritems()))/50 means 1/50 coastline tiles recieve a fish deposit
+		island_fish_ratio = len(sorted(island.ground_map.iteritems()))/50
```
squiddy: This can be shortened to `len(island.ground_map) / 50`. But then, I feel like this should be `len(coastline_tiles) / 50`, the comment says `means 1/50 coastline tiles recieve a fish deposit`.

Or merely the comment is inaccurate, juding by the name `island_fish_ratio`.

horizons/world/worldutils.py
```diff
@@ -258,18 +259,28 @@ def add_nature_objects(world, natural_resource_multiplier):
...
+		
+		for i in xrange(0, island_fish_ratio):
+			tile = coastline_tiles[i*(len(coastline_tiles)/island_fish_ratio)]
+			fish_deposits = 0
+			while fish_deposits == 0:
```
squiddy: A simple

```python
while True:
  ..
  if foobar:
    Build(...)
    break
```

should be enough, no need for the `fish_deposits` variable.

horizons/world/worldutils.py
```diff
@@ -258,18 +259,28 @@ def add_nature_objects(world, natural_resource_multiplier):
...
+				#randomly choose 2 directions to place the fish deposit away from the coastline
+				x_dir = fish_directions[world.session.random.randint(0, 8)][0] + fish_directions[world.session.random.randint(0, 8)][0]
```
squiddy: For the sake of readability, I'd add a shortcut for the random number generation like this:

```python
randint = world.session.random.randint
x_dir = fish_directions[randint(0, 8)][0] + fish_directions[randint(0, 8)][0]
y_dir = fish_directions[randint(0, 8)][1] + fish_directions[randint(0, 8)][1]
...
```

horizons/world/worldutils.py
```diff
@@ -258,18 +259,28 @@ def add_nature_objects(world, natural_resource_multiplier):
...
+				y_dir = fish_directions[world.session.random.randint(0, 8)][1] + fish_directions[world.session.random.randint(0, 8)][1]
+				
+				fish_x = tile.x + x_dir * world.session.random.randint(3, 9)
+				fish_y = tile.y + y_dir * world.session.random.randint(3, 9)
+				if (fish_x, fish_y) in world.ground_map:
+					Build(FishDeposit, fish_x, fish_y, world, 45 + world.session.random.randint(0, 3) * 90, ownerless=True)(issuer=None)
```
squiddy: Shouldn't we check here if the build actually succeeded? Unless I'm missing something here, we could pick a coast tile as start on the east side of the island, but `(x_dir, y_dir)` may lead us the centre of the island where we can not place a deposit.

LinuxDonald:
@squiddy i added some changes from ther pr log here. Are my changes correct? the other needed changes i dont know how to do.

squiddy:
@LinuxDonald I'll look into it tomorrow, see where I can be of help.

PR #2458, opened 2016-08-30
When the cursor is over the editor widget, the cursor changes back to…

LinuxDonald:
… an arrow, rather than a tile.

@MasterofJOKers what do you thinking about that code?

### DIFF
--- a/horizons/editor/gui.py
+++ b/horizons/editor/gui.py
@@ -207,8 +207,8 @@ class SettingsTab(TabInterface):
 		super(SettingsTab, self).__init__(widget=self.widget)
 
 		self._world_editor = world_editor
-		self.current_tile = 'sand'
-		self.ingame_gui = ingame_gui
+		self._current_tile = 'sand'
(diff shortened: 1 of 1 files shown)


MasterofJOKers:
@LinuxDonald The code looks OK to me. It does things like I would do them with my current knowledge. At least as far as I understand the problem and without looking further into the issue. I would probably rename the methods and make them private. And there seems to be a different indentation in the second `mapEvents`. Also, I don't know why the second `mapEvents` is needed, but that might be, because I don't know the GUI/workflow that this commit changes.

LinuxDonald:
Thx for the info. I will test the changes from here and post my feedback.

MasterofJOKers:
Rebased onto current master ...

LinuxDonald:
Tested. Works very good and an nice enhancement :)

@ThePawnBreak thx for this :)

@MasterofJOKers thx for fixing it up :)

PR #2459, opened 2016-08-30
Menu game separation

LinuxDonald:
@MasterofJOKers could you take an loke at this changes? When you say code looks good i will test it.


MasterofJOKers:
> 17:07:33 <gscai> the idea is simple - separate the main menu and the game session into separate processes
17:07:55 <gscai> they would both be controlled by a launcher process
17:08:08 <gscai> the advantages:
17:08:31 <gscai> * recovery from crashes/memory leaks in the game session
17:08:56 <gscai> * might simplify starting using a different gui manager
17:09:19 <gscai> * possible to restart the game/menu as needed; for example to change the resolution
17:09:28 <gscai> the disadvantages:
17:09:42 <gscai> * extra code to achieve the separation
17:10:03 <gscai> the question is whether it is actually worth doing it

We need to look if that's fixed:

> 17:15:58 <gscai> squiddy: it is missing logging support
17:16:11 <gscai> although that could be trivially done by using 1 log per process
17:16:29 <gscai> and it is missing fife support to handle events in the same window
17:17:32 <gscai> + error handling when loading a game fails
17:17:53 <gscai> none of it has to be massive
17:20:03 <gscai> but the strongest advantage is session recovery
17:20:17 <gscai> and it isn't all that strong

MasterofJOKers:
I'll try to re-create the changes on the current master, so that we can have a better discussion about the code.

PR #2467, opened 2016-08-31
ChangeBackground fix & route menu fix & hovereffect fix

CrafterSvK:
**Now you can merge**
Removed hover effect from /horizons/gui/tabs/tradetab.py because of confusion...
Removed other hover text which is not needed now..
Removed picture of hover effect...

ChangeBackground label in mainmenu is now hoverable and you can clearly see it

Not hovered:

Hovered:

Repositioned route menu

snimka obrazovky z 2016-09-01 09-02-33


MasterofJOKers:
Sorry, but that's not how it should be. Pull Requests should be for one specific thing, not changing stuff all over the place. We cannot merge it like that, but need to split it up and test it individually.

Additionally, the commit messages need to be changed this time before merging. I'm looking at you @LinuxDonald

CrafterSvK:
@MasterofJOKers @LinuxDonald told me they will teach me git. Sorry about this. Everything should work. Try it out if you can.

horizons/gui/gui.py
```diff
@@ -78,6 +80,13 @@ def on_escape(self):
 		if self._windows.open_popup(_("Quit Game"), message, show_cancel_button=True):
 			horizons.main.quit()
 
+	def mouse_entered_changebackground(self):
+		changebackground = self._gui.findChild(name='changeBackground')
+		changebackground.background_color = (0, 0, 0, 255)
+
+	def mouse_exited_changebackground(self):
+		changebackground = self._gui.findChild(name='changeBackground')
+		changebackground.background_color = (0, 0, 0, 102)
```
MasterofJOKers: If this is defined in the corresponding XML file, it should be taken from there or saved in `mouse_entered_changebackground` and replaced here. It should not be hard-coded at two different places.  Otherwise, some future developer might be confused why it doesn't work.

MasterofJOKers:
I've merged the changes regarding the "Change Background". @LinuxDonald please have a look at the other changes and tell me, if I should merge them, too.

CrafterSvK:
Closing because of #2470

PR #2473, opened 2016-09-03
Mouse sensitivity [NOT FINISHED]

CrafterSvK:
**Don't merge**
**DONE:**
Changed visual appearance

**TODO:**
Mouse is too fast even on lovest setting
Remove "+"  which  breaks game
Make the scale more normal? Default setting was 0.0. I don't know how to change 0 to 1 without making change to speed of mouse. @MasterofJOKers can you help with these?

### DIFF
--- a/content/gui/xml/mainmenu/settings.xml
+++ b/content/gui/xml/mainmenu/settings.xml
@@ -101,7 +101,7 @@
 			</HBox>
 
 			<Slider size="300,20" name="mousesensitivity" is_focusable="0"
-				  orientation="0" scale_start="0.5" scale_end="5.0" step_length="0.25" />
+				  orientation="0" scale_start="0.1" scale_end="5.0" step_length="0.25" />
 		</VBox>
 	</VBox>
 </HBox>
--- a/content/gui/xml/mainmenu/settings.xml
+++ b/content/gui/xml/mainmenu/settings.xml
@@ -101,7 +101,7 @@
 			</HBox>
 
 			<Slider size="300,20" name="mousesensitivity" is_focusable="0"
-				orientation="0" scale_start="-0.1" scale_end="0.5" step_length="1.0" />
+				  orientation="0" scale_start="0.5" scale_end="5.0" step_length="0.25" />
 		</VBox>
 	</VBox>
 </HBox>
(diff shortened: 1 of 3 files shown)


content/gui/xml/mainmenu/settings.xml
```diff
@@ -101,7 +101,7 @@
 			</HBox>
 
 			<Slider size="300,20" name="mousesensitivity" is_focusable="0"
-				orientation="0" scale_start="-0.1" scale_end="0.5" step_length="1.0" />
+				  orientation="0" scale_start="0.1" scale_end="5.0" step_length="0.25" />
```
squiddy: So this changes the maximum value of the setting internally from 0.5 to 5.0. 

The FIFE code mentions a allowed range: 

https://github.com/fifengine/fifengine/blob/e875cedd88d5860e3f9d03af2b8bf86771e7c07f/engine/core/eventchannel/eventmanager.h#L133

```cpp
/** Sets mouse sensitivity
* The sensitivity is limited to the range -0.99 - 10.0.
*/
void setMouseSensitivity(float sensitivity);
```

Would it make sense to allow the whole range?

content/settings-template.xml
```diff
@@ -19,7 +19,7 @@
 		<Setting name="Lighting" type="int"> 0 </Setting>
 		<Setting name="GLUseFramebuffer" type="bool"> True </Setting>
 		<Setting name="GLUseNPOT" type="bool"> True </Setting>
-		<Setting name="MouseSensitivity" type="float"> 0.0 </Setting>
+		<Setting name="MouseSensitivity" type="float"> 1.0 </Setting>
```
squiddy: I think this should be left at zero. If I understand the code correctly, we're directly passing the value of the slider to FIFE. The setting in FIFE allows an absolute range of [-0.99 - 10.0], so keeping the default at zero shouldn't change the mouse speed.

horizons/gui/modules/settings.py
```diff
@@ -225,7 +225,7 @@ def slider_change(self, widget):
 		value = {
 			'volume_music':      lambda x: u'%s%%' % int(500 * x),
 			'volume_effects':    lambda x: u'%s%%' % int(200 * x),
-			'mousesensitivity':  lambda x: u'%+.1f%%' % (200 * x),
+			'mousesensitivity':  lambda x: u'%+.1f' % (x),
```
squiddy: What's the idea here? Previously the shown value would go from 0 to 100%, is the (rather arbitrary) range of 0 to 5 more easily understood by the player?

LinuxDonald:
I close this because. No feedback from the creator.

PR #2481, opened 2016-09-17
Add .editorconfig file

squiddy:
The editorconfig file has builtin support by a couple editors with many plugins
adding support to others. This should make it easier for people to contribute
since there is no editor configuration necessary.


LinuxDonald:
Thx for this :) :+1:

PR #2483, opened 2016-09-17
Small changes to prepare python 3 support

squiddy:
There is more to do to make UH work with python 3, but these changes should be pretty safe.


squiddy:
What's the minimum python version we're supporting?

MasterofJOKers:
currently? 2.6. or do you mean for python3? I'd say 3.4 as it's in Debian stable.

squiddy:
Yeah, I wanted to know versions from both 2.x and 3.x, thanks.

MasterofJOKers:
On the other hand, 3.5 is in the next Debian stable, so that should be fine, too, if there's something worth it init.

MasterofJOKers:
Looking at the relative import changes: Should we finally start sorting our imports?

squiddy:
Yeah, sorting imports sounds good. isort is pretty capable. We can do this in a different PR, I'm probably going to merge this later, seeing as it won't break anything.

jorgesumle:
>  Imports should be grouped in the following order:
> 
>     standard library imports
>     related third party imports
>     local application/library specific imports
> 
> You should put a blank line between each group of imports. 

Take this from Python style guide (PEP 8) into account: https://www.python.org/dev/peps/pep-0008/#imports.

squiddy:
Added isort in 21c0b64984805795f0416e6c1f7c3029adc1fd14

Issue #2485, opened 2016-09-19
Consider using the nosetests flaky plugin for GUI tests


squiddy:
Since we enabled gui tests on CI, about 10 tests showed inconsistent behaviour, i.e. sometimes they succeed and sometimes not.

We still have about 65 gui tests that succeed, in order to not fail the whole build I marked these inconsistent tests with a `mark_flaky` decorator that will simply mark the test as skipped if it fails. Obviously this is not a long-term solution.

I've briefly looked at https://github.com/box/flaky, which let's us define how many tests reruns have to pass in order to have the test succeed.

squiddy:
Tests seem relatively stable now, we probably don't need that any time soon.

PR #2486, opened 2016-09-19
Enable code coverage tracking with coveralls

squiddy:
Two things happen here. For one, we're collecting coverage data from gui
tests (support initially added in 563f8d4a).
    
Since we're running the unit/integrationtests at first to get early
feedback, we need to set a different COVERAGE_FILE in order to combine
the coverage data with the one from the gui tests.
    
Once everything's done, send the combined coverage data to coveralls.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,3 +2,8 @@
 branch = True
 parallel = True
 source = horizons
+
+include =
+    horizons/*
+    run_uh.py
+    run_server.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,13 @@ install:
   - "cp -a /usr/lib/python2.7/dist-packages/fife/ $VIRTUAL_ENV/lib/python2.7/site-packages/"
   - "prename 's/\\.x86_64-linux-gnu//' $VIRTUAL_ENV/lib/python2.7/site-packages/fife/*.so"
   - "pip install -r requirements.txt"
+  - "pip install coverage coveralls"
 script: 
   - python -c 'from fife import fife; print fife.getVersion()'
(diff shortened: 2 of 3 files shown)


jakoch:
The `.coveragerc` needs to exclude some files from coverage, e.g.
- https://coveralls.io/builds/7955800/source?filename=%2Fhome%2Ftravis%2Fvirtualenv%2Fpython2.7.10%2Flib%2Fpython2.7%2Fsite-packages%2Fnose%2Fplugins%2Fmultiprocess.py
- https://coveralls.io/builds/7955800/source?filename=%2Fhome%2Ftravis%2Fvirtualenv%2Fpython2.7.10%2Flib%2Fpython2.7%2Fsite-packages%2Fyaml%2F__init__.py

Not tested:

```
[report]
omit =  
    */site-packages/nose/*
    */site-packages/yaml/*
    *__init__*
```

squiddy:
@jakoch Good catch. I've defined a whitelist though. I also reactivated the coverage support for gui tests.

LinuxDonald:
Travis is now broken :( https://travis-ci.org/unknown-horizons/unknown-horizons/builds/161447779

squiddy:
Not that it helps much, but the build is broken ever since the gui tests were reactivated. The last 10 commits on master are me trying to figure out which tests are genuinely broken and which are just randomly failing from time to time.

ChrisOelmueller:
I'd much rather have failing tests than disable / not run tests at all because some aren't working :)

squiddy:
The decision to disable some of them was not an easy one, I can assure you. However, I didn't want them to keep failing until someone comes around and fixes them. I'm planning to fix those that are actually broken and for the rest use #2485.

PR #2488, opened 2016-09-22
Reimplement finding widgets in gui tests

squiddy:


MasterofJOKers:
`where the name is C` should be `B` in bae57c6 I guess.

tests/gui/test_helper_find.py
```diff
@@ -0,0 +1,145 @@
...
+
+		helper = GuiHelper(mock.Mock(), mock.Mock())
+		self.find = lambda x: helper._find(self.widgets, x)
+
+	def test_simple(self):
+		"""
+		Single widget in the global list.
+		"""
+		target = self.w('target')
+		assert self.find('target') == target
```
MasterofJOKers: Why not use `self.assertEqual()` or something like that. Just to be sure, that when using `python -O` the tests are still tests.

tests/gui/test_helper_find.py
```diff
@@ -0,0 +1,145 @@
...
+	"""
+	Tests about finding widgets in pychan's internal lists.
+	"""
+	def setUp(self):
+		self.widgets = set()
+
+		def w(name, children=None):
+			widget = Widget(name, children)
+			self.widgets.add(widget)
+			return widget
```
MasterofJOKers: Shouldn't this function also add all children of `widget` into `self.widgets`? Not doing so makes it differ from a real example from what you explained to me.

MasterofJOKers: You're doing it explicitly for every widget, because you also have to create them through this function. Got it.

tests/gui/test_helper_find.py
```diff
@@ -0,0 +1,145 @@
...
+
+		self.w('bla', [
+			self.w('baz', [
+				target
+			])
+		])
+		self.w('foobar')
+		self.w('target')
+
+		assert self.find('bla/target') == target
```
MasterofJOKers: Really? That's supposed to work? I'd expect it to work this way:
```python
target1 = self.w('target')
target2 = self.w('target')
self.w('bla', [ self.w('baz', [ target1 ]) ])
self.w('bla', [ target2 ])

assert self.find('bla/target') == target2
```

Maybe make holes in the path work with a parameter or by using a special char like `/*/`.

You're probably trying to make the test-writing shorter with this, but IMHO it adds unexpected behaviour.

tests/gui/test_helper_find.py
```diff
@@ -0,0 +1,145 @@
...
+		target = self.w('target')
+
+		self.w('wrongpath', [
+			self.w('foobar')
+		])
+		self.w('foobar', [
+			target
+		])
+
+		assert self.find('foobar/target') == target
```
MasterofJOKers: I'm missing tests with multiple children on the same widget. Additionally, I'd like to see a test with a target widget having children, to see that processing children stops at the right time.

tests/gui/helper.py
```diff
@@ -191,40 +190,40 @@ def _get_children(self, w):
...
-		widgets = deque(self.active_widgets)
+		path_components = name.split('/')
 
-		path_components = list(reversed(name.split('/')))
+		first_part = path_components.pop()
+		filtered = [w for w in widgets if w.name == first_part]
+		if not filtered:
+			return None
 
+		candidates = [(f, [f.parent]) for f in filtered]
```
MasterofJOKers: Is my interpretation correct, that a widget can only ever have a single parent? I.e. I cannot use the same widget instance in e.g. a button in two different tabs, but always have to create a new one?

tests/gui/helper.py
```diff
@@ -191,40 +190,40 @@ def _get_children(self, w):
...
+			path = path_components.pop()
+			new_candidates = []
+			for candidate, up in candidates:
+				w = up[-1]
+				while w and w.name != path:
+					w = w.parent
 
-		return match
+				if w and w.name == path:
+					new_candidates.append((candidate, up + [w]))
```
MasterofJOKers: Maybe you should save the whole path to the widget, so that later you can decide to return the result matching exactly the requested path instead of erroring out because of ambiguity.

tests/gui/test_helper_find.py
MasterofJOKers: Having looked at the implementation now, this seems like a bigger problem to me, as I cannot provide a more specific path to `target2`. The only choice would be to rename some part. So at least an `exact=True` should be suppliable to `find()` for cases like this.

tests/gui/helper.py
MasterofJOKers: I'm just asking this, because it seems possible to me, that a widget might be in multiple `children` lists and thus there might be a searched path to it, while with having only a single parent it might be the first/last parent seen by this widget.

tests/gui/test_helper_find.py
squiddy: Old habit, we're using py.test at work. :)

tests/gui/test_helper_find.py
squiddy: You're right, that is an issue.

I don't like the "special character/parameter for holes" solution, because I'm pretty sure that a lot of tests would require this. We're using purely layout related components all over the place, such as VBox or HBox, and they don't get assigned names.

The `exact` parameter sounds okay, I think your idea from another comment is my favourite though:

> Maybe you should save the whole path to the widget, so that later you can decide to return the result matching exactly the requested path instead of erroring out because of ambiguity.

That sounds pretty reasonable and solves the "unexpected behaviour" problem. In the case of multiple matches, `find` would always return the one whose path matches the requested path the best == the candidate with the shortest path of names.

tests/gui/helper.py
squiddy: Looking at pychan's code, lots of things seem to point towards a single parent.

`pychan.widgets.widget.Widget`

```python
		# Parent attribute makes sure we only have one parent,
		# that tests self.__parent - so make sure we have the attr here.
		self.__parent = None
		self.parent = parent

```

When you add a child to a widget (say a Container), you use `addChild` which does

```python
		widget.parent = self
```

But yes, it seems possible that a widget ends up in two `children` lists, however it will only ever have one parent. If it ends up in multiple lists I'd consider it a bug on our side.

squiddy:
I think I've addressed all your points in the update.

tests/gui/test_helper_find.py
```diff
@@ -0,0 +1,224 @@
...
+			target1
+		])
+		self.w('bla', [
+			target2
+		])
+
+		try:
+			self.find('bla/target')
+		except Exception as e:
+			self.assertEqual(str(e), 'Ambigious specification bla/target, found 2 matches')
```
MasterofJOKers: Too bad 2.6 doesn't have `assertRaisesRegexp()`, yet.

squiddy: Yeah, I missed it.

squiddy: I support I could have added a custom Exception class for easier assertion, but I didn't feel like it would be worth it.

tests/gui/helper.py
```diff
@@ -191,40 +190,49 @@ def _get_children(self, w):
...
+
+		if len(candidates) > 1:
+			candidates = sorted(candidates, key=lambda c: len(c[1]))
+			best_matches = [c[0] for c in candidates if len(c[1]) == len(candidates[0][1])]
+
+			if len(best_matches) > 1:
+				raise Exception('Ambigious specification {}, found {} matches'.format(
+					name, len(best_matches)))
+			else:
+				return best_matches[0]
```
MasterofJOKers: I wouldn't use an `else` here, because I see the `if` as validation which breaks if it doesn't validate and the `return` as normal code path. But that's just preference, I guess.

squiddy:
Merged.

PR #2495, opened 2016-10-02
using a dict comprehension to do the same job

MarkusHackspacher:
The change make the code faster and can be use for Python Version 2.7 and above.
I change the passing a list comprehension to the `dict` function (Old style) in a a dict comprehension.

https://www.quantifiedcode.com/app/issue_class/539Wia7V
https://www.python.org/dev/peps/pep-0274/


MasterofJOKers:
merged

PR #2496, opened 2016-10-03
Consider set comprehensions instead of passing lists to 'set()'

MarkusHackspacher:
change set([x for (x, _) in self.land_manager.village]) to {x for (x, _) in self.land_manager.village}

https://www.quantifiedcode.com/app/issue_class/2e5IzmR1

### DIFF
--- a/horizons/ai/aiplayer/areabuilder.py
+++ b/horizons/ai/aiplayer/areabuilder.py
@@ -177,7 +177,7 @@ class AreaBuilder(WorldObject):
 			if not self.settlement_manager.production_builder.road_connectivity_cache.is_connection_possible(collector_coords, destination_coords):
 				return None
 
-		blocked_coords = set([coords for coords in builder.position.tuple_iter()]).union(self.land_manager.coastline)
+		blocked_coords = {coords for coords in builder.position.tuple_iter()}.union(self.land_manager.coastline)
 		beacon = Rect.init_from_borders(loading_area.left - 1, loading_area.top - 1,
 		                                loading_area.right + 1, loading_area.bottom + 1)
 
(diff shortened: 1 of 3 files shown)


MasterofJOKers:
I'm not sure about what we decided, but python2.6 does not seem to support that type of syntax.
```python
>>> a = [1, 2, 3, 3]
>>> {x for x in a}
  File "<stdin>", line 1
    {x for x in a}
         ^
SyntaxError: invalid syntax
>>> set(x for x in a)
set([1, 2, 3])
```

MasterofJOKers:
@LinuxDonald told me, we're not supporting python2.6 any more. So that's fine.

I'd still prefer the `set()` over the `{}` as it's more explicit. But that's probably more because of me not using `{}` much.

MarkusHackspacher:
I have only one argument, it is faster.
But the set() it easier to read. Maybe help inline comments.

```
>>> import timeit
>>> timeit.timeit('{x for x in [1, 2, 3, 4]}')
0.5454318523406982
>>> timeit.timeit('set(x for x in [1, 2, 3, 4])')
0.8383378982543945
>>> timeit.timeit('{x for x in [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]}')
1.1321148872375488
>>> timeit.timeit('set(x for x in [1, 2, 3, 4, 5, 6, 7, 8, 9, 0])')
1.6722140312194824
```

LinuxDonald:
Maybe add an comment for easier understanding in the code?

MasterofJOKers:
No comment necessary to explain syntax.

You're right. It's about 200 ns faster on my system. So if that function is called very very often, that would make a difference.

LinuxDonald:
@MasterofJOKers is that ready for merge? And can that pushed too into python3 branch?

MasterofJOKers:
@LinuxDonald Not by you, no. @ChrisOelmueller said something about it on IRC regarding the removal of the line, that seems to do nothing. It should go in its own commit. So ... I could probably merge it if I wouldn't be playing right now.

MasterofJOKers:
merged

PR #2497, opened 2016-10-03
Remove polib from ext

MarkusHackspacher:
The polib library is only use in the development/create_pot.sh file. And a old copy is in horizons/ext/polib.py. Version 1.0.3, actual is 1.0.7 https://pypi.python.org/pypi/polib and supports any version of python ranging from 2.4 to latest 3.X version.

In the other files like run_tests.py are a copy from a used library not in the ext, only it the requirements.txt.
And if not installed it make a message:

```
try:
    import nose
except ImportError:
    print('The nose package is needed to run the UH tests.')
    sys.exit(1)
```

I remove the polib from ext and change the development/create_pot.sh consistent to the others.


MasterofJOKers:
This might be something, where @ChrisOelmueller's opinion would be nice, as he has most knowledge about translations afaik.

LinuxDonald:
That shouldn't breaking anything. It uses than the system polib.

MasterofJOKers:
There might be a reason not to do so - bugs for example or differing behaviour, which in turn invalidates all translations or something like that.

ChrisOelmueller:
In this case, I don't recall bugfixes being the reason for including `polib` but rather easy access to what is needed for translation-related scripts. "The system polib" might not be available (like, what would that even be on Windows), at which point I'd much prefer my scripts still working as opposed to having to install stuff from pypi manually. What's the motivation behind removing it?

Unrelated, I'd try to have one commit for "remove one file", right now it looks more like five(?).

MasterofJOKers:
Making commits great again is the task of the person merging the PR, I guess.

MasterofJOKers:
Following that argument, we'd need to include every library not in the python standard library into the project source - wouldn't we?

MasterofJOKers:
After discussion on the IRC: I'm going to merge this later/tomorrow.

PR #2498, opened 2016-10-04
`else` clause on a loop without a `break` statement

MarkusHackspacher:
have add a break statement.
All tests are fine, but I found no test for only this unit.

The else clause of a loop is executed when the loop sequence is empty. When a loop specifies no break statement, the else clause will always execute, because the loop sequence will eventually always become empty. Sometimes this is the intended behavior, in which case you can ignore this error. But most times this is not the intended behavior, and you should therefore review the code in question.

```
def contains_magic_number(list, magic_number):
    for i in list:
        if i == magic_number:
            print "This list contains the magic number."
    else:
        print "This list does NOT contain the magic number."

contains_magic_number(range(10), 5)
# This list contains the magic number.
# This list does NOT contain the magic number.

# add a break statement 
def contains_magic_number(list, magic_number):
    for i in list:
        if i == magic_number:
            print "This list contains the magic number."
            break  # added break statement here
    else:
        print "This list does NOT contain the magic number."

contains_magic_number(range(10), 5)
# This list contains the magic number.
```
https://www.quantifiedcode.com/app/issue_class/4aqWoDeY

### DIFF
--- a/horizons/gui/mousetools/tearingtool.py
+++ b/horizons/gui/mousetools/tearingtool.py
@@ -88,6 +88,7 @@ class TearingTool(NavigationTool):
 				self.session.view.renderer['InstanceRenderer'].removeColored(building._instance)
 				if (not building.id in BUILDINGS.EXPAND_RANGE) or self.confirm_ranged_delete(building):
 					Tear(building).execute(self.session)
+					break
 			else:
 				if self._hovering_over:
 					# we're hovering over a building, but none is selected, so this tear action isn't allowed
--- a/horizons/gui/mousetools/tearingtool.py
+++ b/horizons/gui/mousetools/tearingtool.py
@@ -84,11 +84,11 @@ class TearingTool(NavigationTool):
 				self.coords = coords
 			self._mark(self.coords, coords)
 			selection_list_copy = [building for building in self.selected]
-			for building in selection_list_copy:
-				self.session.view.renderer['InstanceRenderer'].removeColored(building._instance)
-				if (not building.id in BUILDINGS.EXPAND_RANGE) or self.confirm_ranged_delete(building):
-					Tear(building).execute(self.session)
-					break
+			if self.selected:
+				for building in selection_list_copy:
+					self.session.view.renderer['InstanceRenderer'].removeColored(building._instance)
+					if (not building.id in BUILDINGS.EXPAND_RANGE) or self.confirm_ranged_delete(building):
+						Tear(building).execute(self.session)
 			else:
 				if self._hovering_over:
 					# we're hovering over a building, but none is selected, so this tear action isn't allowed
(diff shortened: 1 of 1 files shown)


MasterofJOKers:
464ab939b56d82d33c7d34634d048ddfa9969f94 is the commit introducing the `else`. Since the code is essentially the same, my guess would be that the better fix is to remove the `else`.

Your code would only tear down the first building of a selection, if I see that correctly. So inserting the `break` does not seem to be the best idea.

From what I understand, the best fix would be the following, as it seems to be the intent of the original author.
```python
if self.selected:
  for i in self.selected:
    […]
else:
  if self._hovering_over:
    […]
```

MasterofJOKers:
Now we need @LinuxDonald to test if there's a difference in the behaviour and if that difference is ok.

squiddy:
> All tests are fine, but I found no test for only this unit.

Why not write a test then? It can serve as a check whether the change broke something, and most times is a good way to become familiar with a particular piece of code.

Using code coverage, we can also check if this code is covered by any test at all: https://coveralls.io/github/unknown-horizons/unknown-horizons

MarkusHackspacher:
A test write by people who are not so deep in the structure inside has certain advantages and brings by another approach new insights. But the source text is not commented, ok maybe two lines. Is quite hard.
Helpful in the file is a `assert isinstance(session, horizons.session.Session)` like in
https://github.com/MarkusHackspacher/unknown-horizons/blob/master/horizons/gui/mousetools/cursortool.py#L41
And a copy of at horizons.session.Session in a existing testcase.

And unified function names
http://pylint-messages.wikidot.com/messages:c0103

LinuxDonald:
This pr works. I dont become any ingame message about an error. :+1:

horizons/gui/mousetools/tearingtool.py
```diff
@@ -84,10 +84,11 @@ def mouseReleased(self, evt):
...
-				self.session.view.renderer['InstanceRenderer'].removeColored(building._instance)
-				if (not building.id in BUILDINGS.EXPAND_RANGE) or self.confirm_ranged_delete(building):
-					Tear(building).execute(self.session)
+			if self.selected:
+				for building in selection_list_copy:
+					self.session.view.renderer['InstanceRenderer'].removeColored(building._instance)
+					if (not building.id in BUILDINGS.EXPAND_RANGE) or self.confirm_ranged_delete(building):
+						Tear(building).execute(self.session)
 			else:
 				if self._hovering_over:
```
squiddy: These two lines can be merged to `elif self._hovering_over`

MarkusHackspacher:
Good idea, I have fix this.

LinuxDonald:
With this pr you become only the messange that warehouse could not deleted when you click directly on it with the destroy tool. For this is an good thing. 👍

PR #2499, opened 2016-10-05
default arguments should not use a list

MarkusHackspacher:
Mutable list used as default argument to a function. Python creates a single persistent object and uses it for every subsequent call in which the argument is left empty. This can cause problems if the program was expecting the function to return a new list after every call.

https://www.quantifiedcode.com/app/issue_class/3P0qV6OB

### DIFF
--- a/horizons/util/python/decorators.py
+++ b/horizons/util/python/decorators.py
@@ -109,7 +109,8 @@ def temporary_cachedmethod(timeout):
 
 globals().update(opmap)
 
-def _make_constants(f, builtin_only=False, stoplist=[], verbose=False):
+def _make_constants(f, builtin_only=False, stoplist=None, verbose=False):
+	stoplist = stoplist or []
 	try:
 		co = f.func_code
 	except AttributeError:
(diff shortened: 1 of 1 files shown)


horizons/util/python/decorators.py
```diff
@@ -109,7 +109,8 @@ def __call__(self, *args, **kwargs):
 
 globals().update(opmap)
 
-def _make_constants(f, builtin_only=False, stoplist=[], verbose=False):
+def _make_constants(f, builtin_only=False, stoplist=None, verbose=False):
+	stoplist = stoplist or []
```
MasterofJOKers: I'd really prefer the check for `None`, as this this type of setting would use a new list instance instead of the given instance for an empty list. Even though that's probably not the best way to do it, one might want the original instance to be used (e.g. to see in tests, if items were added).

MarkusHackspacher:
In the same file was exactly the same contruct, some lines under my change:
Maybe somebody changes this and looks ok, but late not make the file consistent.
https://github.com/unknown-horizons/unknown-horizons/blob/master/horizons/util/python/decorators.py#L218-L229

MasterofJOKers:
touché.

Then I'll merge your changes and then add a commit myself to make it look like I want to.

MarkusHackspacher:
Ok, that is fine.

PR #2502, opened 2016-10-08
add in all mousetool a instance check for the Session class

MarkusHackspacher:
And add in the class comment a param session.

Have try to write a test by my self, but the test for the TearingTool is already in tests/gui/test_mousetools.py

### DIFF
--- a/horizons/gui/mousetools/attackingtool.py
+++ b/horizons/gui/mousetools/attackingtool.py
@@ -29,15 +29,18 @@ from horizons.gui.mousetools.selectiontool import SelectionTool
 
 
 class AttackingTool(SelectionTool):
-	"""
-		This will be used when attacking units are selected
-		it will have to respond on right click and change cursor image when hovering enemy units
+	"""This will be used when attacking units are selected
+	it will have to respond on right click and change cursor image when hovering enemy units
+
+	@type session: horizons.session.Session
+	@param session: session type
 	"""
 
 	send_hover_instances_update = False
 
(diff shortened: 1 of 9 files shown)


MasterofJOKers:
I would not use `assert()` for anything, that should be still there in production, i.e. on the user system, as `assert()` is not executed when running python with `-O` to optimize.

MasterofJOKers:
What's the purpose of this? I don't see the point in doing that.

MarkusHackspacher:
I saw this in https://github.com/unknown-horizons/unknown-horizons/blob/master/horizons/gui/mousetools/cursortool.py#L41
and make this in all mousetools files.

MasterofJOKers:
That's from this commit 13ecfd88e0fd1603f1fc5ad8950e439190d181a8 In the commit, there is no description of why the assert was added. I think that needs discussion, as checking types in the `__init__()` should be done everywhere or nowhere I guess. This checking is imho against the principle of duck-typing and might hinder testing.

MasterofJOKers:
eoc on the IRC wrote:
> +1 for removing rather than adding that stuff
> on the other hand, it does indeed hinder testing, hmmm

squiddy:
> This checking is imho against the principle of duck-typing and might hinder testing.

100 times this. I'm against more `isinstance` checks.

MasterofJOKers:
Thank you for your proposal, but we don't want to go in that direction. Feel free to create a new pull-request removing the asserts.

PR #2509, opened 2016-10-11
Show UH logo as window icon during atlas creation

squiddy:
Tested on ArchLinux using awesome v3.5.9. Anyone else could test this? Especially windows would be nice.


MasterofJOKers:
hm ... got a question about this: When running a realeased version, there's no need to regenerate the atlas, right? So this is only for devs and git-head-users? I'm just asking this, because when installing from a package, the location for the atlas is not writable by an ordinary user, as it's somewhere in /usr/share, IIRC.

squiddy:
That's a very good point.

You know, initially I only checked whether the icon for the game itself worked, and it did. Read a comment about the atlas thingy not showing an icon, and thought "time to read up on Tk".
After the first problem encountered I was going to drop it because... who actually cares?

You basically answered that, I think devs and git-head-users can do without an icon for the popup. I'm closing this and the issue then.

PR #2510, opened 2016-10-12
Avoid unnecessary name repetition in equality comparisions

MarkusHackspacher:
unnecessary name repetition.

### DIFF
--- a/development/networkclient.py
+++ b/development/networkclient.py
@@ -246,7 +246,7 @@ try:
 except (ValueError, IndexError):
   port = 0
 
-if None in [host, port] or port <= 0:
+if host is None or port is None or port <= 0:
   usage()
   sys.exit(1)
 
--- a/development/networkclient.py
+++ b/development/networkclient.py
@@ -246,7 +246,7 @@ try:
 except (ValueError, IndexError):
   port = 0
 
-if host == None or port == None or port <= 0:
+if None in [host, port] or port <= 0:
   usage()
   sys.exit(1)
 
--- a/horizons/ai/aiplayer/villagebuilder.py
+++ b/horizons/ai/aiplayer/villagebuilder.py
@@ -781,7 +781,7 @@ class VillageBuilder(AreaBuilder):
 
 		removed_coords_list = []
 		for coords, (purpose, (section, _)) in self.plan.iteritems():
-			if purpose == BUILDING_PURPOSE.RESERVED or purpose == BUILDING_PURPOSE.NONE:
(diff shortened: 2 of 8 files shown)


development/networkclient.py
```diff
@@ -246,7 +246,7 @@ def onhelp(*args):
 except (ValueError, IndexError):
   port = 0
 
-if host == None or port == None or port <= 0:
+if None in [host, port] or port <= 0:
```
squiddy: I don't like that one to be honest. The previous code should be improved as well, by using `is` for comparison. It should be

```python
if host is None or port is None
```

horizons/ai/aiplayer/villagebuilder.py
```diff
@@ -781,7 +781,7 @@ def handle_lost_area(self, coords_list):
 
 		removed_coords_list = []
 		for coords, (purpose, (section, _)) in self.plan.iteritems():
-			if purpose == BUILDING_PURPOSE.RESERVED or purpose == BUILDING_PURPOSE.NONE:
+			if purpose in [BUILDING_PURPOSE.RESERVED, BUILDING_PURPOSE.NONE]:
```
squiddy: :+1:

horizons/component/componentholder.py
```diff
@@ -180,6 +180,6 @@ def get_component_template(cls, component):
 		for entry in cls.component_templates:
 			if isinstance(entry, dict):
 				for key, value in entry.iteritems():
-					if cls.class_mapping[key] == component or key == component:
+					if component in [cls.class_mapping[key], key]:
```
squiddy: This feels a bit weird, since you're switching the operands of the comparison. Which is syntactically fine of course, but in this case it hurts readability in my opinion. We're iterating over the items in `entry`, and check whether the key equals the component. This change feels like doing `2 == x` instead of `x == 2`.

run_server.py
```diff
@@ -105,7 +105,7 @@ def usage(fd=sys.stdout):
 except (ValueError, IndexError):
 	port = 0
 
-if host == None or port == None or port <= 0:
+if None in [host, port] or port <= 0:
```
squiddy: Forgot this one, this should be using `is`. See my comment on `development/networkclient.py`.

development/networkclient.py
MarkusHackspacher: I have change.

horizons/component/componentholder.py
MarkusHackspacher: OK, yes the old is better, I change it back.

squiddy:
@MarkusHackspacher Thanks for your contribution.

PR #2512, opened 2016-10-13
Make render name of icons more unique to prevent crash

### LINKED ISSUE #2475: KeyError: 'produced_resource_5_Point(106, 125)_222'
@janexx reported this error to me:
https://gist.github.com/janexx/b20f9cb81727d9c2b58f9c804ae0d537

I found the same crash log here:
https://github.com/unknown-horizons/unknown-horizons/issues/1916

but this should fixed by this:
https://github.com/unknown-horizons/unknown-horizons/pull/1810

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37566963-keyerror-produced_resource_5_point-106-125-_222?utm_campaign=plugin&utm_content=tracker%2F287396&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F287396&utm_medium=issues&utm_source=github).
</bountysource-plugin>

squiddy:
Fixes #2475 

---

I guess we could safely drop at least one other parameter of `get_resource_string`, but I suppose it can't hurt to keep it?

### DIFF
--- a/horizons/world/managers/productionfinishediconmanager.py
+++ b/horizons/world/managers/productionfinishediconmanager.py
@@ -19,6 +19,8 @@
 # 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 # ###################################################
 
+import uuid
+
 from fife import fife
 
 import horizons.globals
@@ -152,4 +154,4 @@ class ProductionFinishedIconManager(object):
 		This key MUST be unique!
 		"""
 		return "produced_resource_" + str(res) + "_" + str(instance.position.origin)\
-		       + "_" + str(Scheduler().cur_tick)
+		       + "_" + str(uuid.uuid1())
--- a/tests/gui/ingame/test_bugs.py
+++ b/tests/gui/ingame/test_bugs.py
@@ -21,12 +21,12 @@
 
 from horizons.command.building import Tear
 from horizons.command.unit import CreateUnit
-from horizons.constants import BUILDINGS, UNITS, RES
+from horizons.component.storagecomponent import StorageComponent
+from horizons.constants import BUILDINGS, RES, UNITS
 from horizons.messaging import ResourceProduced
 from horizons.world.production.producer import Producer
 from tests.gui import gui_test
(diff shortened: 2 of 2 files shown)


horizons/world/managers/productionfinishediconmanager.py
```diff
@@ -152,4 +154,4 @@ def get_resource_string(self, instance, res):
 		This key MUST be unique!
 		"""
 		return "produced_resource_" + str(res) + "_" + str(instance.position.origin)\
-		       + "_" + str(Scheduler().cur_tick)
+		       + "_" + str(uuid.uuid1())
```
MasterofJOKers: Has this any consequences for multiplayer? Shouldn't, right? Just asking, because this takes the current time. Don't know if the scheduling ticks are synced, though.

And reading the documentation, you might want to add parameters to `uuid1()`.

> Generate a UUID from a host ID, sequence number, and the current time. If node is not given, getnode() is used to obtain the hardware address. If clock_seq is given, it is used as the sequence number; otherwise a random 14-bit sequence number is chosen.

https://docs.python.org/2/library/uuid.html#uuid.getnode <- says, that it might be slow and I don't see, how the MAC address is relevant here. You could also use the `Scheduler().cur_tick` as sequence number, I guess.

squiddy: Thing is, and that's what I was getting at it with my question, if I understand it correctly, this value can be any random string, as long as we get no collisions. So even time in nanoseconds would be enough. Probably with the "produced_resource" prefix, but that's it.

This is only used locally for FIFE.

MasterofJOKers: I'd rather user `uuid.uuid4()` then: "Generate a random UUID." The "produce_resource" prefix and the resource might help in debugging, so I would keep that.

squiddy: I've changed it.

Issue #2482, opened 2016-09-17
Discuss code style checks


squiddy:
As someone that hasn't contributed to this project in years, it's easier for me to run a tool that will yell at me when code doesn't adhere to the style guide.

I guess we used to run the `pep8` tool, because there are some settings in `setup.cfg` related to it. However, c289a5173e744bf8f61adc8758ee7d982971e4c4 added a (seemingly more complete) configuration for pylint.

Unfortunately, running `pylint horizons` spits out thousands (26483 for me) of warnings and errors (whitespace issues, indentation, variable names...). I was going to suggest running code style checks in CI, but at this point it's just not feasible.

I guess we have two options:

1) fix warnings and/or adjust code style configuration
2) remove .pylintrc

Personally I'd vote for option 1.

jorgesumle:
We could start with trailing whitespaces:
I've run this: `find . -type f -name '*.py' -exec sed --in-place 's/ \+$//' {} \+`.
This only works for Python .py files. Change the extension to make it work for other formats. I'm not sure if that's necessary.

jakoch:
What about adding it to `travis.yml`:

```
install:
    # Install pylint
    - sudo apt-get install pylint
script:
    # Run pylint
    - pylint --errors-only --rcfile=".pylintrc" horizons/
```

You could set it to report only - but don't let the build fail. 
That should produce a log with things to fix and you can work through it step-by-step.

MarkusHackspacher:
In my fork, I use in travis.yml this:
`  - python pylint-recursive.py`
The output are https://travis-ci.org/MarkusHackspacher/unknown-horizons/builds/160466819
That was also the reason, that I ask for delete the file `development/networkclient.py ` see #2487

squiddy:
We already have a `.pylintrc` in the master branch, so I'm not sure what you're actually suggesting here.

Can you explain what `pylint-recursive.py` is for? If I run `pytest` myself, it will already recurse into subdirectories.

The main issue to me here is that with the current configuration, pylint attempts to do a lot of static analyzing, which fails with the codebase. For example, we probably should configure pylint to ignore the Enum class (there's lots of cases where pylint complains).

Perhaps we should drop the static analyzing for now and concentrate on consistent code style - which still is a lot to do.

squiddy:
@MarkusHackspacher 

> We already have a .pylintrc in the master branch, so I'm not sure what you're actually suggesting here.

[...]

PR #2514, opened 2016-10-15
add missing whitespace around arithmetic operator

MarkusHackspacher:
http://stackoverflow.com/questions/18171560/does-pep-require-whitespace-around-operators-in-function-arguments

### DIFF
--- a/horizons/util/python/decorators.py
+++ b/horizons/util/python/decorators.py
@@ -97,12 +97,11 @@ def temporary_cachedmethod(timeout):
 					del self.cache_dates[key]
 					return self(*args, **kwargs)
 			else:
-				self.cache_dates[key] = time.time() # new entry
+				self.cache_dates[key] = time.time()  # new entry
 
 			return super(_temporary_cachedmethod, self).__call__(*args, **kwargs)
 
-	return functools.partial( _temporary_cachedmethod, timeout=timeout )
-
+	return functools.partial(_temporary_cachedmethod, timeout=timeout)
 
 
 # adapted from http://code.activestate.com/recipes/277940/
@@ -135,7 +134,7 @@ def _make_constants(f, builtin_only=False, stoplist=None, verbose=False):
 		if opcode in (EXTENDED_ARG, STORE_GLOBAL):
 			return f    # for simplicity, only optimize common cases
 		if opcode == LOAD_GLOBAL:
-			oparg = newcode[i+1] + (newcode[i+2] << 8)
+			oparg = newcode[i + 1] + (newcode[i + 2] << 8)
 			name = co.co_names[oparg]
 			if name in env and name not in stoplist:
 				value = env[name]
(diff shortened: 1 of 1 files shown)


horizons/util/python/decorators.py
```diff
@@ -97,12 +97,11 @@ def __call__(self, *args, **kwargs):
 					del self.cache_dates[key]
 					return self(*args, **kwargs)
 			else:
-				self.cache_dates[key] = time.time() # new entry
+				self.cache_dates[key] = time.time()  # new entry
```
squiddy: We only have a handful of lines with inline comments that are prefixed with two spaces. So this and the other change below need to go.

MarkusHackspacher: In the file are already inline comments with more then one space:
https://github.com/unknown-horizons/unknown-horizons/blob/master/horizons/util/python/decorators.py#L117
https://github.com/unknown-horizons/unknown-horizons/blob/master/horizons/util/python/decorators.py#L136

squiddy: Even more so, using more than two spaces as well. So what's the reasoning behind using two spaces then?

Before I wrote the comment, I did a simple check across the codebase. We have far more inline comments with one space than two. So in my opinion, that's the way to go.

MarkusHackspacher: Python's official style guide, [PEP 8](http://www.python.org/dev/peps/pep-0008/), is very clear about this issue:
```
Inline comments should be separated by at least two spaces from the statement.
They should start with a # and a single space.
```
UH use PEP 8, but not everything implemented, see point
 -  **String formatting** should prefer the Python3-style format() over the % operator

squiddy: Actually, we recommend PEP-8 as style guide, but we don't enforce it (that's why I opened #2482). At least not according to the Wiki. And I'm pretty sure we've had the discussion about inline comments before.

While we're at it, allow me me to quote PEP-8 too:

> A style guide is about consistency. Consistency with this style guide is important. Consistency within a project is more important. Consistency within one module or function is the most important.

https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

You're arguing for **Consistency with this style guide is important**, whereas I actually looked at the rest of the codebase to see that (apparantly) most inline comments use one space. You keep ignoring that statement.

squiddy: Previous discussion about inline comments: https://github.com/unknown-horizons/unknown-horizons/pull/2290

MarkusHackspacher: Thank you for find the old statement from @ChrisOelmueller. I think I make it better for code reading with a good example, but this end with

> I don't see the point in arguing about this longer.

In my opinion, we should talk about it a bit. Even if it is annoying for you. *joke
I not ignoring that statement, in my first answer was the are two other line the agrees with the pep8,  in my first was make the file consistent.

> rest of the codebase to see that (apparantly) most inline comments use one space

I think logical: consistent = all the same = true,
most all the same = not consistent = false

squiddy: I'm not going to comment here further. I requested changes. If you apply them, I'm happy to merge this.

If you really feel this needs more discussion, even though 2! years ago we already had that discussion and made our (two long term maintainers) point clear, open an issue. But I won't tolerate you sneaking your style changes into unrelated pull-requests.

MarkusHackspacher:
I remove second space before inline comment.

squiddy:
Thank you.

Issue #2513, opened 2016-10-14
Introduce type hints for static type checking


squiddy:
[PEP-484](https://www.python.org/dev/peps/pep-0484/) introduced new syntax to declare type hints. The interpreter makes no use of them, other then to expose them on e.g. function objects.

Third party tools like [mypy](http://mypy-lang.org/) understand the syntax and can check programs statically.

Note: this won't affect the game at runtime at all. It won't run faster, it won't run slower. It will however most likely help catching bugs and making lovers of static typing happy.

If I understand it correctly, mypy will only check function bodies if we start providing type hints, however running it against our codebase produces some errors. Failure to import FIFE and some complaints about our class inheritance trees. 

- [ ] Fix issues raised when running mypy on clean codebase
- [ ] Identify parts of codebase most likely to benefit the most (I'm thinking of essentials, Scheduler and stuff) and add type hints
- [ ] Profit?
- [ ] Nice to have: add mypy checks to CI

@ChrisOelmueller COMEFROM .idle

MasterofJOKers:
I've read in an article about mypy, that one should start with getting '--check-untyped-defs' running and then continue to add explicit typing to the rest of the code, once this runs without errors and is added to the CI.

squiddy:
Indeed. The article in question (I believe it's [the one from Tulip](http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/)), says to start off with `mypy --py2 --silent-imports --fast-parser -i <paths>`. Once that works, one should add `--check-untyped-defs` what you mentioned.

I've updated the issue accordingly.

squiddy:
I'd say we skip checking the tests for now, especially in GUI tests a lot of monkey-patching is going on.

squiddy:
I'm working on the first step and I'm wondering how we should handle the following problem (so far I haven't found clues in the documentation):

```python
from typing import TYPE_CHECKING, Any

if TYPE_CHECKING:
	from horizons.util.uhdbaccessor import UhDbAccessor

# The global Fife instance. This is used both inside a game session and in the menus.
fife = None # type: Any

# The global read-only data database (includes most content/*.sql)
db = None # type: UhDbAccessor
```

This is `horizons/globals.py` with type hints. Even when using comments, one needs to import types. Running this would error since `typing` is not installed. An easy solution could be to ship `typing` with UH, but that doesn't sound right.

MasterofJOKers:
[...]

PR #2517, opened 2016-10-15
Configuration for mypy and some initial type hints

squiddy:
See #2513 

For the most part these changes are pretty straightforward, there is one issue however that I described in https://github.com/unknown-horizons/unknown-horizons/issues/2513#issuecomment-253968927

Some type hints require importing from the `typing` module, which we need to either add to the dependencies, ship with our code or do something entirely different.

### DIFF
--- a/doc/LICENSE
+++ b/doc/LICENSE
@@ -43,6 +43,10 @@ obtained at http://www.opensource.org/licenses/mit-license.php .
 is licensed by Armin Ronacher under the terms of the 3-clause BSD license. See
 BSD_speaklater.
 
+/horizons/ext/typing.py
+
+is licensed under the Python License, Version 2. See https://github.com/python/typing
+
 
 The files
 
--- a/horizons/ai/aiplayer/basicbuilder.py
+++ b/horizons/ai/aiplayer/basicbuilder.py
@@ -21,11 +21,10 @@
 
 import copy
 
-from typing import Tuple
-
 from horizons.command.building import Build
 from horizons.constants import BUILDINGS
 from horizons.entities import Entities
+from horizons.ext.typing import Tuple
 from horizons.util.python import decorators
 from horizons.util.shapes import Point, Rect
 from horizons.world.building.production import Mine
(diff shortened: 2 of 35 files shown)


squiddy:
I might work on this during the next week, but with the exception of one issue this is good to merge: https://github.com/unknown-horizons/unknown-horizons/issues/2513#issuecomment-253968927

However we could certainly take the approach @MasterofJOKers suggested here https://github.com/unknown-horizons/unknown-horizons/issues/2513#issuecomment-253991882

```python
import my
import normal
import stuff

if False:
    from typing import Any
    from some.package.somewhere import Somethnig
```

squiddy:
Mind you, there are still errors left to fix. We're still in phase 1 of the mypy master plan.

ChrisOelmueller:
I'd be okay with `if False` or rather, something like this:

```py
if False: # TYPE_CHECKING
    from typing import Any
    from some.package.somewhere import Something
```

squiddy:
I think we've come to a solution regarding the imports: https://github.com/unknown-horizons/unknown-horizons/issues/2513#issuecomment-254494723

So I'll extend this PR by doing the first 2 points, then I could merge it.

squiddy:
I'd like to merge this as it is. It adds some type hints, but there are still many things to prepare/fix. But this serves as foundation for future work.

squiddy:
This takes a bit longer, because now that we're including typing in horizons/ext, mypy insists on checking that file, which produces warnings.

LinuxDonald:
Are here open questions? Or could it be merged?

squiddy:
My comment right above yours states that there is an open issue.

LinuxDonald:
Ah okay sorry.

squiddy:
Merged in 0d876feeb568c2f7dc95325b6960b8830247837b.

It is unfortunate that currently mypy doesn't support excluding specific files or directories from checks, however that shouldn't stop us from adding more type hints. Errors regarding `horizons.ext` can always be ignored (grep -v).

I was tempted to go back to the `if False` approach, but mypy would still run against horizons.ext. We'll figure this out later.

Issue #2515, opened 2016-10-15
Remove uses of Dummy in horizons package


squiddy:
mypy has an issue with the magic Dummy code (who doesn't?):

> horizons/ext/dummy.py:63: error: Invalid assignment target

I was going to move it to the tests package, since that's where we need it. However, we have a handful of places where we use that code inside horizons, which is a bad idea really, so:

- [ ] Move horizons.ext.dummy to tests
- [ ] Fix imports of Dummy in tests (mostly setup I think)
- [x] Find alternative to Dummy, or better, refactor the now broken parts in horizons to not rely on such hacks

PR #2521, opened 2016-10-16
No dummy in minimap

MasterofJOKers:
I tried to do one of the points of #2515. Any thoughts?

### DIFF
--- a/horizons/gui/widgets/minimap.py
+++ b/horizons/gui/widgets/minimap.py
@@ -95,6 +95,8 @@ class Minimap(object):
 		@param on_click: function taking 1 argument or None for scrolling
 		@param preview: flag, whether to only show the map as preview
 		@param tooltip: always show this tooltip when cursor hovers over minimap
+
+		NOTE: Preview generation in a different process overwrites this method.
 		"""
 		if isinstance(position, Rect):
 			self.location = position
--- a/horizons/gui/widgets/minimap.py
+++ b/horizons/gui/widgets/minimap.py
@@ -21,7 +21,6 @@
 from __future__ import print_function
 
 import itertools
-import json
 import math
 from math import cos, sin
 
--- a/horizons/gui/widgets/minimap.py
+++ b/horizons/gui/widgets/minimap.py
@@ -35,6 +35,78 @@ from horizons.util.python import decorators
 from horizons.util.shapes import Circle, Point, Rect
 
 
+def get_world_to_minimap_ratio(world_dimensions, minimap_dimensions):
+	"""Compute the number of pixels of the world needed for one pixel on the minimap.
+
+	Returns a tuple for x and y, in case they differ.
+	Accepts two tuples of (width, height) to compute the ratio on.
+	"""
+	return tuple(float(w) / m for w, m in zip(world_dimensions, minimap_dimensions))
+
+
+def iter_minimap_points(location, world, island_color, water_color, area=None):
+	"""Return an iterator over the pixels of a minimap of the given world.
+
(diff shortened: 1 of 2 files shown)


squiddy:
I think you are on a good way with `Move point calculations of minimap in own method`, however I don't really like subclassing `Minimap` and completely replacing its `__init__` method.

Without looking too closely at the code, couldn't we extract the `_iter_points` method into a single function, that gets everything that it needs and returns the data?

`Minimap` could use that function in `_recalculate`, and the preview code merely needs to call that function, no class needed, e.g.:

```python
data = get_minimap_data(island, something, else)
data = [(point.x, point.y, r, g, b) for point, (r, g, b) in data]
print(json.dumps(data))
```

MasterofJOKers:
I'll give it a try. There's more than one function that needs extraction, though. Something about the world_minimap_ratio is cached on the instance.

squiddy:
If it seems to be too much work, it's probably not worth it. Your changes are already a nice improvement over using Dummy, so we can keep it like that as well. We're only using this for the preview anyway. We can always iterate on it at some later time.

MasterofJOKers:
In the end, only two functions had to be extracted, because the preview does not use rotation. The rotation is done for every point in `Minimap._recalculate()`.

horizons/gui/widgets/minimap.py
```diff
@@ -465,90 +534,34 @@ def cleanup():
...
 
+		for (x, y), color in iter_minimap_points(self.location, self.world,
+						self.COLORS["island"], self.COLORS["water"], where):
+			if use_rotation:
+				# inlined _get_rotated_coords
+				rot_x, rot_y = self._rotate((location_left + x, location_top + y), self._rotations)
+				fife_point.set(rot_x - location_left, rot_y - location_top)
+			else:
+				fife_point.set(x, y)
+			fife_point.set(x, y)
```
squiddy: This looks wrong to me, I can't find an equivalent to this in the old code. You're already setting the point in the condition above.

horizons/gui/widgets/minimap.py
```diff
@@ -36,6 +35,78 @@
 from horizons.util.shapes import Circle, Point, Rect
 
 
+def get_world_to_minimap_ratio(world_dimensions, minimap_dimensions):
+	"""Compute the pixel of the world needed for one pixel on the minimap.
```
squiddy: Maybe "Compute *the number of pixels* of the world..."?

horizons/gui/widgets/minimap.py
MasterofJOKers: Yeah, that seems totally wrong. It would essentially cancel out rotation, thus making it the way we discussed on IRC we might want to make it.

MasterofJOKers:
Made the changes and gave it a try: Preview generates as expected, but the in-game minimap does not rotate. @LinuxDonald is that, what you meant with "it is broken"? Because then it's not my fault :grin:

squiddy:
Can confirm, minimap doesn't rotate on master already. For me, this is good to go.  :shipit:

LinuxDonald:
The minimap rotaing is broken for long time. I get not broken with this pr ;)

MasterofJOKers:
Even if the setting for it to rotate is enabled? I've just had a looked and it's disabled for me.

MasterofJOKers:
So if I enable the setting, the minimap rotates when I click it. Default seems to be off.

LinuxDonald:
Okay I don't knowed that there is an setting for... But the minimap view is different to the map view

Issue #2466, opened 2016-08-31
Game sometimes failed to exit


CrafterSvK:
Sometimes is game unquitable. I am running it with terminal so i can terminate i iwth clicking the cross.

LinuxDonald:
Yeah i know that bug but i dont know why i happens... I have it often too when i want to close uh :(

squiddy:
I've noticed something similar, didn't really happen before. I'm using awesome WM, and when running UH from the terminal, I can interrupt it with `CTRL+C`.

However, when I use the awesome WM shortcut `CTRL+SHIFT+C` to kill the current window, it just freezes (i.e. can't move the cursor inside the game) and I'm no longer able to interrupt on the terminal.

LinuxDonald:
For me it sometimes freeze when i click on the X in window mode.

squiddy:
I'm pretty sure this is caused by locks in threads, with the preloading thread I can reproduce it. If I start the game, and shortly after seeing the menu try to end it (either by shortcut or clicking X), it will hang. `strace` reports this:

`futex(0x558d5704f400, FUTEX_WAIT_BITSET_PRIVATE|FUTEX_CLOCK_REALTIME, 0, NULL, ffffffff`

I can get the process to hang repeatedly. If however I disable the preloading with `python run_uh.py --no-preload` this issue is not present. 

My guess is the lock isn't released when the application is instructed to close.

squiddy:
Reopened, because this still happens when inside the editor.

LinuxDonald:
I had the issue still to when uh crashes.

squiddy:
> I had the issue still to when uh crashes.

Sorry, but I don't understand what you're trying to say. You can't close the game after it crashed?

LinuxDonald:
I can't close it after it crashed. It hangs then.

squiddy:
Does this happen after every crash? And what crash exactly? So far we only
have two documented crashes.

Am 27.10.2016 12:09 vorm. schrieb "Thomas Kowaliczek" <
notifications@github.com>:

squiddy:
Interestingly enough, when I start the editor from the command line `run_uh.py --edit-map full-house` and wait until the game loaded I can exit the game just fine. If however I'm starting the editor from the main menu, exiting will hang the game.

PR #2524, opened 2016-10-16
Fix one possible cause of hangups when closing the game window

squiddy:
I don't know why, but when closing the game by either clicking the
windows' close button (think windows or ubuntu) or using a shortcut of
the window manager, fife dispatches 2 quit commands.

This will trigger `horizons.main.quit` two times, which in turn will
attempt to join the preloading thread two times. Before joining, the
code in `preload_game_join` attempts to acquire the lock, which will
cause the hangup. The reason why the join function needs to acquire the
lock first is another mistery to me.

Regardless, we shouldn't handle the two quit commands. Once we initiated
shutdown, we can ignore the second command.

Refs #2466

### DIFF
--- a/horizons/gui/keylisteners/mainlistener.py
+++ b/horizons/gui/keylisteners/mainlistener.py
@@ -97,5 +97,8 @@ class MainListener(fife.IKeyListener, fife.ICommandListener, LivingObject):
 
 	def onCommand(self, command):
 		if command.getCommandType() == fife.CMD_QUIT_GAME:
-			horizons.main.quit()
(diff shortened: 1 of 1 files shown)


squiddy:
@LinuxDonald Could you test this branch, see if you still get hangups when closing the window? I tried and it seems good for now, at least in the main menu after start.

LinuxDonald:
Seem that it is now fixed :) good work :+1:

squiddy:
Landed in https://github.com/unknown-horizons/unknown-horizons/commit/b8bbbc30bc303cfd13d17f64f1e102532ad39144

PR #2503, opened 2016-10-08
splitting assert statements and further use from assert horizons.session.Session

MarkusHackspacher:
I take a look to the assert statements, I split it for a better diagnose.

And check where is use the assert horizons.session.Session with git grep horizons.session.Session.
In the file world/**init**.py have I activated and got this If I click at random map, for a new game.
Or at [Travis CI](https://travis-ci.org/MarkusHackspacher/unknown-horizons/builds/166106795):
python2 run_tests.py tests.gui.ingame.test_windows:test_settings_dialog_crash2 -a gui
python2 run_tests.py tests.gui.menu.test_multiplayer:test_create_game -a gui
python2 run_tests.py tests.gui.menu.test_singleplayer:test_start_map -a gui

```
 File "/unknown-horizons/horizons/gui/modules/singleplayermenu.py", line 425, in _update_map_preview
    world = load_raw_world(map_file)
 File "/unknown-horizons/horizons/world/__init__.py", line 741, in load_raw_world
    world = World(session=None)
 File "/unknown-horizons/horizons/world/__init__.py", line 86, in __init__
    assert isinstance(session, horizons.session.Session)
AttributeError: 'module' object has no attribute 'session'
```

PR #2526, opened 2016-10-18
Use pep8 for trailing whitespace check 

MarkusHackspacher:
The trailing whitespace remove my IDE automatically, according to the .editorconfig settings.
But that has influence if I want to edit other, these are also removed. See: https://github.com/unknown-horizons/unknown-horizons/pull/2503#discussion_r82510052

The idea was make a check where in the project all the trailing whitespace.

### DIFF
--- a/tests/unittests/test_pep8.py
+++ b/tests/unittests/test_pep8.py
@@ -28,7 +28,7 @@ import pep8
 
 
 class TestCodeFormat(TestCase):
-	"""automatical test for code format use pep8""" 
+	"""automatical test for code format use pep8"""
 
 	def test_pep8_dir_tests(self):
 		"""Test that code conform to PEP8 test all files"""
--- a/tests/unittests/test_pep8.py
+++ b/tests/unittests/test_pep8.py
@@ -36,7 +36,7 @@ class TestCodeFormat(TestCase):
 			'W2',    # whitespace warnings
 		]
 
-		checker = pep8.StyleGuide(select=tests, paths=['.'], reporter=pep8.StandardReport)
+		checker = pep8.StyleGuide(select=tests, paths=['horizons', 'tests', 'development'], reporter=pep8.StandardReport)
 		report = checker.check_files()
 
 		self.assertEqual(report.total_errors, 0,
--- a/tests/unittests/test_pep8.py
+++ b/tests/unittests/test_pep8.py
@@ -21,7 +21,6 @@
 # 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 # ###################################################
 
-import os
 from unittest import TestCase
 
 import pep8
--- a/tests/unittests/test_pep8.py
+++ b/tests/unittests/test_pep8.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+
+# ###################################################
+# Copyright (C) 2008-2016 The Unknown Horizons Team
+# team@unknown-horizons.org
+# This file is part of Unknown Horizons.
+#
+# Unknown Horizons is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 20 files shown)


tests/unittests/test_pep8.py
```diff
@@ -0,0 +1,45 @@
...
+# 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+# ###################################################
+
+import os
+from unittest import TestCase
+
+import pep8
+
+
+class TestCodeFormat(TestCase):
```
squiddy: I don't think this should be part of the test suite, but instead be integrated into the CI flow on travis. The configuration you added in the test can probably go into either setup.cfg or .pep.rc or whatever it's called, and contributers can then run `pep8` to run the style checks.

MarkusHackspacher: My idea is to solve the problems, and not make the thing complicated. If I have brought with this PR someone to another or better idea then it helps the project. And my PR can then be closed.

But I would be happy if it was accepted, maybe even only until there is something better. Is only one file which is then deleted.

squiddy: @MarkusHackspacher I thought about it a bit more, let's find a compromise: If you could fix the issues the test shows in this PR, I'm willing to merge it.

MasterofJOKers:
If I might chime in: We also have/had a pre-commit hook, that every developer was supposed to be running. Imho that would be a good place for such a check. (Maybe additionally to the CI. I would not expect it in the Tests.)

squiddy:
> Maybe additionally to the CI

Doesn't matter how, but the style checks must run on CI and especially for PRs. We can't rely on people setting up the hooks. I can imagine an issue with a hook: what if someone "broke" master and I just want to work on my branch, am I blocked from comitting/pushing then until someone fixes master?

MasterofJOKers:
The my hook at work works, is that certain checks are only done on the files or even changes to be commited. So no, you would not be blocked.

MarkusHackspacher:
I´m ready, I have remove the trailing whitespace.

tests/unittests/test_pep8.py
```diff
@@ -0,0 +1,45 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the
+# Free Software Foundation, Inc.,
+# 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+# ###################################################
+
+import os
```
squiddy: This import seems unused.

tests/unittests/test_pep8.py
```diff
@@ -0,0 +1,45 @@
...
+	"""automatical test for code format use pep8"""
+
+	def test_pep8_dir_tests(self):
+		"""Test that code conform to PEP8 test all files"""
+		tests = [
+			# http://pep8.readthedocs.org/en/latest/intro.html#error-codes
+			'W2',    # whitespace warnings
+		]
+
+		checker = pep8.StyleGuide(select=tests, paths=['.'], reporter=pep8.StandardReport)
```
squiddy: I think we need to be more explicit here. This basically runs pep8 on everything in the current working directory, yes?

This should specify horizons/, tests/, development/ explicitly. Otherwise I think this pep8 check might run on packages I have in my virtualenv for example.

MarkusHackspacher:
I change your requested points and hope the best for your check today evening.

LinuxDonald:
@squiddy coudl this get merged? or should @MasterofJOKers review it too?

squiddy:
@LinuxDonald I'm looking into it right now.

squiddy:
So I found out, that we already have a pep8 config in setup.cfg, added by @MarkusHackspacher . I've modified that to start with a clean config and added the check in the CI workflow. 

Then I've cherry picked the commit with the whitespaces fixes from here. We can add more pep 8 rules step by step.

MarkusHackspacher:
In the moment I cannot say it good or not. The travis CI pep8 test is not running. You should add pep8 in  requirements.txt, you can cherrypick here:
https://github.com/unknown-horizons/unknown-horizons/pull/2526/commits/f3f804217ae1a5ffe49e9215ab10d7f7713d93be

Issue #2518, opened 2016-10-15
Simplify horizons.util.color by dropping metaclass magic


squiddy:
While experimenting with mypy (see #2513), I got the following error:

```
horizons/util/color.py: note: In member "__iter__" of class "ColorMeta":
horizons/util/color.py:51: error: Generic type is prohibited as a runtime expression (use a type alias or '# type:' comment)
```

I can't claim to understand that, and a related issue (python/typing#79) doesn't help, but the code seems overly complicated.

The main object here is the `Color' class. Certain default colors are stored in the database with an id and name.

```python
c = Color(0, 0, 255) # create a new color instance
c = Color(255, 255, 255) # if found in the DB, gets assign an id and name

for c in Color:
  print(c) # iterate over all colors in the database

c = Color['red'] # returns a new instance of Color using name with values from the database
c = Color[0] # returns a new instance of Color using id with values from the database
```

By dropping the metaclass, we would loose the last two features. However, they can easily reimplemented with classmethods. It could look like this:

```python
for c in Color.defaults:
    print(c)

c = Color.by_name('red')
c = Color.by_id(0)
# or even
c = Color.get(0)
c = color.get('red')
```

squiddy:
Another thought: do we even need colors in the database? We might as well hardcode them in my opinion.

MasterofJOKers:
Do we need all the other stuff in the database?

squiddy:
I don't know. I was asking in the context of simplying the color code, not questioning the whole database.

Some of the stuff in `game.sql` is nice to have outside the source, e.g. related_buildings, resources. If a database is the right fit is a different question.

Those 14 colors haven't changed in at least 4 years now, I don't see them changing in the future. But we can keep them in there, this issue primarily is about cleaning up the code.

MasterofJOKers:
I don't get the whole database thing for reading only simple values. Could as well load it from yaml, as far as I am concerned. Maybe some values are filtered on a specific column. That would be the only reason for me. On the other hand, one could still create a database out of yaml, if one has to do sql queries.

PR #2527, opened 2016-10-20
Remove metaclass magic from Color

squiddy:
This can be reimplemented in a similar style, but without the magic.

Closes: #2518


LinuxDonald:
@MasterofJOKers please review it :)

squiddy:
@LinuxDonald I'm going to remove the milestone from this PR, because the underlying issue #2518 is already assigned to that milestone. 

Otherwise https://github.com/unknown-horizons/unknown-horizons/milestone/20 would show 2 entries for basically the same thing.

LinuxDonald:
It's okay :-)

PR #2537, opened 2016-10-27
remove whitespace after and before '(' Part 1

MarkusHackspacher:
I remove in some files the whitespace after and before '('.


horizons/network/server.py
```diff
@@ -62,30 +62,30 @@ def __init__(self, hostname, port, statistic_file=None):
...
+			packets.client.cmd_creategame:     [self.oncreategame],
+			packets.client.cmd_listgames:      [self.onlistgames],
+			packets.client.cmd_joingame:       [self.onjoingame],
+			packets.client.cmd_leavegame:      [self.onleavegame],
+			packets.client.cmd_chatmsg:        [self.onchat],
+			packets.client.cmd_changename:     [self.onchangename],
+			packets.client.cmd_changecolor:    [self.onchangecolor],
+			packets.client.cmd_preparedgame:   [self.onpreparedgame],
+			packets.client.cmd_toggleready:    [self.ontoggleready],
+			packets.client.cmd_kickplayer:     [self.onkick ],
```
squiddy: There is a trailing whitespace after `self.onkick`

horizons/component/depositcomponent.py
```diff
@@ -44,11 +44,11 @@ def initialize(self, inventory=None):
...
+		return ((res, data.get('min_amount', 0), data['max_amount'])
+		        for res, data in self.resources.iteritems())
 
 	def get_random_res_amounts(self):
 		"""Generator for tuples (res_id, rand_amount) for each resource that the deposit
 		can contain (as defined in the object file)."""
-		return ( (res, self.session.random.randint(min_amount, max_amount))
-		         for res, min_amount, max_amount in self.get_res_ranges() )
+		return ((res, self.session.random.randint(min_amount, max_amount))
+		        for res, min_amount, max_amount in self.get_res_ranges())
```
ChrisOelmueller: So now I have to wonder, do you truly think the two lines are more readable this way?

MarkusHackspacher: Yes, but that is taste. What do you think?

MarkusHackspacher:
There was another error, I have already delete the rule in the select config. Have 323 case found, this is too large for one commit, so I have call this part 1.

horizons/component/depositcomponent.py
squiddy: I think it didn't change the readability, the original version isn't really good to start with, it's still confusing what's really happening there. I suppose a more explicit version with a for loop and yield would be way easier to read.

Overall this is a good change, so I'm going to merge it.

Issue #2451, opened 2016-08-29
Scenario default language


CrafterSvK:
When you try to play scenario it selects first language available not the one you set in options. Would be great if it selects english as default or the one you selected in options when available.
##

MasterofJOKers:
I cannot reproduce that. It works for me, even if I change the language, while the game is running.

LinuxDonald:
For me it works too no problems. @CrafterSvK could you maybe recheck this?

CrafterSvK:
It works for languages that have translated tutorial. If you have not it will select Catalana or something like that and than it doesn't recheck on the other scenario and keeps Catalana... (Sorry but I am ill. So i can't concentrate on what I am typing)

MasterofJOKers:
That's a much better explanation. Thank you.

squiddy:
@CrafterSvK Hi, I've submitted a patch with modified behaviour here: https://github.com/unknown-horizons/unknown-horizons/commit/3bc51f602a0f54fcf38fea9cdc035d196b409819 Perhaps you can try it out. :-)

LinuxDonald:
I have tested it a bit. For me it works. That language that i select in the options get selectetd to in the scenario by default

PR #2543, opened 2016-10-30
Improve language selection in scenario menu

squiddy:
When switching between scenarios and both support the currently selected
language, the selection will be kept.

When switching to a scenario that doesn't support the currently selected
language, we try to determine the next best one by looking at the system
locale. When selecting another map, we once again first try to use the
language from the settings.

Also rewrite some helper functions to be stateless.

Fixes: #2451


LinuxDonald:
:+1: for merge

PR #2545, opened 2016-10-30
Use git describe for version detection

squiddy:
mainmenu-version
### TODO
- [x] Test git describe on windows

### DIFF
--- a/content/gui/xml/mainmenu/mainmenu.xml
+++ b/content/gui/xml/mainmenu/mainmenu.xml
@@ -3,7 +3,7 @@
 	<Icon image="content/gui/images/background/mainmenu/wheel.png" position="212,44" />
 	<Icon image="content/gui/images/logos/fife.png" position="775,20" />
 
-	<Label name="version_label" text="Unknown Horizons" position="735,735" />
+	<Label name="version_label" text="Unknown Horizons" position="40,685" />
 	<Label name="changeBackground" text="Change Background" position="755,685" />
 
 	<!-- clockwise, starting at 0 degree -->
--- a/horizons/constants.py
+++ b/horizons/constants.py
@@ -24,6 +24,7 @@ import ctypes
 import os
 import os.path
 import platform
+import subprocess
 
 from horizons.ext.enum import Enum
 from horizons.ext.typing import Optional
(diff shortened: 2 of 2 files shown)


squiddy:
Tested successfully under Windows 7 Enterprise, Python 2.7.12, git 2.10.1.

LinuxDonald:
Before we pull this pr we should make new tag for the version.

squiddy:
I don't understand how that is related. This extends the version detection by adding the tag name, if available. Nothing more.

squiddy:
Also this code isn't relevant for releases, only for git checkouts. Releases get their version number hardcoded.

LinuxDonald:
Okay

jorgesumle:
Works fine on GNU/Linux. I'm running _Trisquel 7.0_.

LinuxDonald:
👍 For merge

krichter722:
Why hasn't this been merged? It's useful for orientation in development builds. Please start using git tags.

ChrisOelmueller:
It was merged, cf. e75c9ae2c26a33a040b1cf3afec9c6baf46be33e

LinuxDonald:
Our lastest git tag is version 2014.x because since this version we dont had yet released an new one. The next will be 2017.1 but i cant say you yet an date.

krichter722:
> It was merged, cf. e75c9ae

Good to know. Thank you.

> Our lastest git tag is version 2014.x because since this version we dont had yet released an new one. The next will be 2017.1 but i cant say you yet an date.

I see the tags in `git tag`, but `git describe` says `fatal: No tags can describe 'e7e701c9334a8680c1ad96fe09ba7f52eddd6dd0'.`. I usually have this issue if tags are used on branch different from master (e.g. if a projects decides to do so) (see also http://stackoverflow.com/questions/4154485/git-describe-ignores-a-tag).

We can discuss this in another issue if you want.

ChrisOelmueller:
That happens because the (to me, questionable) default of `git describe` is to disregard tags that are not annotated. Include the `--tags` flag to change that behavior and also find light (un-annotated) tags, which I'd expect the majority of recent tags to be.

krichter722:
> That happens because the (to me, questionable) default of git describe is to disregard tags that are not annotated. Include the --tags flag to change that behavior and also find light (un-annotated) tags, which I'd expect the majority of recent tags to be.

`git describe --tags` fails due to the same error.

krichter722:
>> That happens because the (to me, questionable) default of git describe is to disregard tags that are not annotated. Include the --tags flag to change that behavior and also find light (un-annotated) tags, which I'd expect the majority of recent tags to be.
>
> git describe --tags fails due to the same error. I assume git fetch --tags fetches all types of tags.

Works fine now as of 2014.1-1013-g9797b7127 (possibly earlier).

PR #2546, opened 2016-10-30
E201 whitespace after '(' Part 2

MarkusHackspacher:
Hallo, the next part of E201 whitespace after '('.


LinuxDonald:
👍 For merge

horizons/ai/aiplayer/productionbuilder.py
```diff
@@ -263,24 +263,24 @@ def display(self):
 			return
 
 		tile_colors = {
-			BUILDING_PURPOSE.ROAD:            ( 30,  30,  30),
```
squiddy: Is there a way to exclude blocks of code from certain checks? I prefer the original layout, especially when lots of lines are aligned in a tabular-like style.

MarkusHackspacher: Yes, I find it look very nice with the table layout, too.
But this break some code style rules. And it could be color, but I am not sure.
But if it a color maybe it is possible make a hexadecimal color definition or real names?
I think hexadecimal is one step better, real names colors two steps.

https://de.wikipedia.org/wiki/Hexadezimale_Farbdefinition

ChrisOelmueller: Is there a way to exclude blocks of code from certain checks?

squiddy: That wasn't what I asked though.

If we were to switch to `flake8`, we can at least skip style rules for individual lines.

squiddy: @MarkusHackspacher I think we should continue thinking for ourselves and not blindly following other projects without any thought.

I'm confused by your example. There is additional whitespace, or am I going crazy?

```python
"grey_light"    :   pygame.Color(200, 200, 200),
```

And by the way, I didn't find any mention of either `pep8`, `pycodestyle`, `pyflakes` or `flake8` in that repo. Looking at the library code, I seriously doubt they do any style checks at all.
Perhaps that should inspire us.

MarkusHackspacher: Yes, this example break the rule: E203 whitespace before ‘:’ and E241 multiple spaces after ':'.
To make it conform, it look like this:
```
            BUILDING_PURPOSE.ROAD: (30, 30, 30),
            BUILDING_PURPOSE.FISHER: (128, 128, 128),
            BUILDING_PURPOSE.LUMBERJACK: (30, 255, 30),
            BUILDING_PURPOSE.TREE: (0, 255, 0),
            BUILDING_PURPOSE.FARM: (128, 0, 255),
            BUILDING_PURPOSE.POTATO_FIELD: (255, 0, 128),
            BUILDING_PURPOSE.PASTURE: (0, 192, 0),
            BUILDING_PURPOSE.WEAVER: (0, 64, 64),
            BUILDING_PURPOSE.SUGARCANE_FIELD: (192, 192, 0),
```

ChrisOelmueller: Could you maybe re-read this thread from the beginning and follow the other side's statements and arguments a bit more closely?

MarkusHackspacher: Sorry.
> Is there a way to exclude blocks of code from certain checks?

I see only to exclude the complete file in the docs.
--exclude=patterns   exclude files or directories
https://pycodestyle.readthedocs.io/en/latest/intro.html#configuration

> I prefer the original layout, especially when lots of lines are aligned in a tabular-like style.

For my opinion is the tabular-like style good for to look fast what is happen a row under, but in this it is not important, what values in the other row. On the contrary, it brings the reader confused because he believes in the first moment that would means something. But each line stands for itself.

LinuxDonald: i like too the tabular-like style. I think its eaysier for reading but all other changes could be merged.

MarkusHackspacher: Another proposal, the values written as hexadecimal?
```
		BUILDING_PURPOSE.ROAD:            (0x1e, 0x1e, 0x1e),
		BUILDING_PURPOSE.FISHER:          (0x80, 0x80, 0x80),
		BUILDING_PURPOSE.LUMBERJACK:      (0x1e, 0xff, 0x1e),
		BUILDING_PURPOSE.TREE:            (0x00, 0xff, 0x00),
```

ChrisOelmueller: Hm. So leading zeroes would be okay, but leading spaces aren't? The hex values are more readable or helpful than the numeric ones?

There's also this little problem:

```py
>>> 030 == 30
False
```

MarkusHackspacher: Yes, this idea to save the tabular-like style was not so good. Leading zero is a octal number and hex not good readable. But is a color code `#0088FF` like HTML not possible?
```
>>> print(030)
24
```

ChrisOelmueller: So what would you suggest?

To find out what is and isn't possible, we have to check the code that's actually using these numbers. Which line(s) would that be here?

MarkusHackspacher: I found in fife.py this line (14918)
If it possible to make the same for hex, for example: `def addHexColored(self, *args):`?
```
    def addColored(self, *args):
        """
        addColored(InstanceRenderer self, Instance instance, int32_t r, int32_t g, int32_t b, int32_t a=128)
        addColored(InstanceRenderer self, Instance instance, int32_t r, int32_t g, int32_t b)
        """
        return _fife.InstanceRenderer_addColored(self, *args)
```

LinuxDonald:
Any update here about the merge status?

horizons/util/pathfinding/pathfinding.py
```diff
@@ -193,14 +193,13 @@ def execute(self):
 				                 and i not in blocked_coords )
 			else:
 				# all relevant vertical and horizontal neighbors
-				neighbors = ( i for i in ((x-1, y), (x+1, y),
+				neighbors = (i for i in ((x-1, y), (x+1, y),
 				                          (x, y-1), (x, y+1) )
```
ChrisOelmueller: Could you also adjust the spacing in this line to match the changes, please?

horizons/gui/style.py
```diff
@@ -26,16 +26,16 @@
...
          255: Color(80, 80, 40, 255),
 }
 
-WHITE = { 32: Color(255, 255, 255,  32),
-          64: Color(255, 255, 255,  64),
-          96: Color(255, 255, 255,  96),
+WHITE = {32: Color(255, 255, 255,  32),
+         64: Color(255, 255, 255,  64),
+         96: Color(255, 255, 255,  96),
          160: Color(255, 255, 255, 160),
```
ChrisOelmueller: Please exclude the changes in this file from the PR, we'd like to keep in the old state for code layout reasons.

horizons/ai/aiplayer/villagebuilder.py
```diff
@@ -841,13 +841,13 @@ def display(self):
...
 			BUILDING_PURPOSE.RESIDENCE:      (255, 255, 255),
-			BUILDING_PURPOSE.ROAD:           ( 30,  30,  30),
+			BUILDING_PURPOSE.ROAD:           (+30,  30,  30),
 			BUILDING_PURPOSE.VILLAGE_SCHOOL: (128, 128, 255),
 			BUILDING_PURPOSE.PAVILION:       (255, 128, 128),
 			BUILDING_PURPOSE.TAVERN:         (255, 255,   0),
 			BUILDING_PURPOSE.FIRE_STATION:   (255,  64,  64),
 			BUILDING_PURPOSE.DOCTOR:         (255, 128,  64),
-			BUILDING_PURPOSE.RESERVED:       (  0,   0, 255),
+			BUILDING_PURPOSE.RESERVED:       (000,   0, 255),
```
ChrisOelmueller: Please exclude the changes in this file from the PR, we'd like to keep in the old state for code layout reasons.

horizons/ai/aiplayer/productionbuilder.py
```diff
@@ -263,24 +263,24 @@ def display(self):
...
-			BUILDING_PURPOSE.CLAY_PIT:        (  0,  64,   0),
-			BUILDING_PURPOSE.BRICKYARD:       (  0,  32,   0),
+			BUILDING_PURPOSE.CLAY_PIT:        (000,  64,   0),
+			BUILDING_PURPOSE.BRICKYARD:       (000,  32,   0),
 			BUILDING_PURPOSE.BOAT_BUILDER:    (163,  73, 164),
 			BUILDING_PURPOSE.SALT_PONDS:      (153, 217, 234),
-			BUILDING_PURPOSE.HERBARY:         ( 64, 200,   0),
-			BUILDING_PURPOSE.RESERVED:        (  0,   0, 128),
+			BUILDING_PURPOSE.HERBARY:         (+64, 200,   0),
+			BUILDING_PURPOSE.RESERVED:        (000,   0, 128),
```
ChrisOelmueller: Please exclude the changes in this file from the PR, we'd like to keep in the old state for code layout reasons.

horizons/ai/aiplayer/productionbuilder.py
ChrisOelmueller: Let's just keep the old layout instead of messing with stuff like that. I believe we got enough opinions from the team to support that decision.

ChrisOelmueller:
Merged in https://github.com/unknown-horizons/unknown-horizons/commit/edffef915a755e832ffc12ed53dd86f3a0c439d3

PR #2567, opened 2016-11-13
Stance dont merge.

LinuxDonald:


LinuxDonald:
Merged with master. But maybe needs first Autoadjustcontainer from pychan_port branch!

PR #2572, opened 2016-11-13
Profile game for a while without taking loading work into account

LinuxDonald:


LinuxDonald:
@MasterofJOKers or @squiddy could one from you maybe take an look on this if we could need this please?

squiddy:
This is not a change that is ever going to be merged into master. It's more like a code snippet. So why don't we just leave it where it is?

Sorry to be so blunt, but to me, all these pull-requests only do one thing well: remove focus from the upcoming release. Some of these branches exist for over 4 years, if they stayed around for one more month it wouldn't matter.
Then there is the question of whether it's wise to merge year-old changes this close to a release.

LinuxDonald:
I don't want to merge them yet. Because I don't want break anything. But after release they can maybe get merged.

PR #2578, opened 2016-11-25
Ambient graphics

LinuxDonald:


janexx:
Ah thanks @LinuxDonald  :)
But maybe we must optimize a bit: 

- the ambient graphics are on beach too, I think they shouln't be there?
- in my opinion the ambient graphics are too much, maybe they should appear a bit less?
- the treestumps should appear only when trees were cutted down (at least it was @ChrisOelmueller's  proposal)

MasterofJOKers:
[ ] Don't put the ambient graphics on the beach.

I don't think treestumps should only appear, if trees were cut down, because trees might also have fallen in a storm or something like that. They might not appear very ofthen, though. Also, when would be the right time to remove a treestump? Only "when something is built over it" might leave a lot of treestumps around a lumber jack's hut if trees are planted at random. That might not look that good.

janexx:
You are right @MasterofJOKers Maybe the treestumps can randomly appear after cutting a tree down, so that just sometimes a stump appears when a tree was cutted down. The most trees could disappear after cutting them down likte it is now. In addition a few treestumps can appear because of other reasons (storm, old age,...) 

Or the treestumps can just appear completety randomly at the beginning of the game.

LinuxDonald:
@melloskitten maybe that is something for you?

melloskitten:
Sir yessir! One question though: is this here solely for the tree-stump issue or what do the others mean by ambient graphics?

LinuxDonald:
You could fix up the code from this pr. It needs to be ported to python3 and it's needs to the fix that it appears on beach. And the ideas about the tree stumps needed added to.

LinuxDonald:
Branch is now uptodate with master. it works with python3.
Todos:
The same as for the tree fix ambient dont should be on sand tile.
Add the ideas about tree stumps.

melloskitten:
thanks! i will get to that asap.

LinuxDonald:
Thx for the feedback :)

LinuxDonald:
@melloskitten any news on this?

melloskitten:
Sorry bro, have a lot of stuff to do for uni + thesis, but i have it in the back of my head and i will finish it :)

squiddy:
Let's break this up, in this PR:

* Cleanup code
* Have stumps appear randomly at start of game
* No stumps on the beach

The idea regarding cut trees can be implemented later.

LinuxDonald:
@melloskitten how it looks about your time now? :)

melloskitten:
@LinuxDonald Actually it looks a lot better now! I handed in my thesis last week so im a free elf now! haha 
i'll take a look at this asap, since i have to set up my environment again & get fife, fifechan etc working since i had to switch to ubuntu

LinuxDonald:
@melloskitten any news here?

LinuxDonald:
@melloskitten are you still working on this?

jmdejong:
Good news: the changed code is python independent, so it works on python3 without fixing
Preventing ambient objects on beaches is very easy, the same check as for trees can be used.
I have done this in PR #2809.

About the graphics: I think the mushrooms are too bright and too large.
A group of mushrooms is already almost half as the size of a deer.
Personally I would prefer to see some lower bushes, and more variations of green/brown/yellow vegetation

LinuxDonald:
Branch is now rebased

jmdejong:
What shall we do with this?
Do other agree that the mushrooms are too large? Should I disable them for now?
Or do you prefer to keep it like this?

Does the code need refactoring or other changes? (currently `check_file_for_tree` checks for ambient things as well)

LinuxDonald:
@janexx could you maybe make the mushrooms smaller please?

janexx:
Ok I will do that.

Issue #2583, opened 2016-12-14
Update handling of FIFE exceptions


Niektory:
A recent FIFE patch (https://github.com/fifengine/fifengine/pull/945) changed the types of exceptions thrown to Python clients. Previously they were all converted to `RuntimeError`, while now they have custom types such as `fife.NotFound`. Because of this they are no longer caught properly by Unknown Horizons.

I see two ways to fix it:
- Update all exception handlers to the new types. This is difficult because it would require determining the correct type in each case which is not always clear from the code.
- Monkey-patch `fife.Exception` to inherit from `RuntimeError`. This is a bit hacky but it would allow old handlers to work unchanged (or be updated selectively), while new code could use the new exception types.

I'd go with the second option.

One other thing: Should it be implemented in a way that it works with both old and new FIFE versions, or is it not necessary and only new versions need to be supported?

LinuxDonald:
Only new version because next uh release will need current fife release.

LinuxDonald:
I like the second option. But i think when we hack/workaround this bug we should let open this ticket or open new one to remove (for lone time turn) the hack fix.

Issue #2516, opened 2016-10-15
Remove __builtin__ modification for gettext


squiddy:
Right now, we're doing a single `gettext.install('de', ...)` once in the code, and the rest of the modules magically have access to `_`, `_N` and other things.

This has some issues:
- danger of local shadowing by using `_` as "unused argument", e.g. when unpacking
- need to configure linters to ignore "not defined" errors when looking at `_`
- mypy doesn't understand it, because builtins aren't really supposed to be modified #2513

My suggestion would be:
- Expose `_` etc. on `horizons.i18n` and import it explicitly, like this:

``` python
from horizons.i18n import ugettext as _

_('Foobar')
```

We already do that for lazy translations, e.g. `from horizons.i18n import _lazy`.

The advantage when using `install` is, that all code continues to work after changing a language, and `_` always points to the active translations's `gettext`. I know the Django project has solved that somehow, you just `from django.utils.translation import gettext as _` and it works.

---

As a first step, I'd be fine with keeping the install, but adding the imports mentioned above everywhere we use `_`, `_N` etc.
- [x] Expose `_` & co. in `horizons.i18n`, so that `from horizons.i18n import ugettext as _, ugettext_lazy as _lazy` works.
- [x] Replaces uses of _ in files with imports

MasterofJOKers:
I was wondering, how that worked. :+1: for changing it. "explicit is better than implicit"

Niektory:
I seems `run_uh.py` was left out and needs `from horizons.i18n import gettext as _` too. Not sure where exactly to put it considering this warning: https://github.com/unknown-horizons/unknown-horizons/blob/master/run_uh.py#L49

PR #2584, opened 2016-12-16
Update handling of FIFE exceptions

Niektory:
Issue #2583


horizons/gui/tabs/boatbuildertabs.py
```diff
@@ -163,14 +164,10 @@ def update_queue(self, container_active):
 
 			try:
 				icon = Icon(name=icon_name, image=image, helptext=helptext)
-			except RuntimeError,e:
+			except fife.NotFound,e:
```
MasterofJOKers: I'd really like this to change to `fife.NotFound as e` or *at least* add a space between the ',' and the 'e'.

run_uh.py
```diff
@@ -355,6 +355,12 @@ def setup_fife():
 	else:
 		log().debug('Using fife version %s.%s.%s, at least %d.%d.%d required', fife_version_major, fife_version_minor, fife_version_patch, VERSION.REQUIRED_FIFE_MAJOR_VERSION, VERSION.REQUIRED_FIFE_MINOR_VERSION, VERSION.REQUIRED_FIFE_PATCH_VERSION)
 
+	# FIFE 0.4.0 changed the types of exceptions from RuntimeError to FIFE-specific types
+	# this line is needed for old FIFE exception handlers to work
+	# when all handlers are updated to use the new types it can be removed
+	# see issue #2583
+	fife.Exception.__bases__ = RuntimeError,
```
MasterofJOKers: This looks really very hacky. Are there that many exception handlers left or what's the problem with changing the remaining handlers?

MasterofJOKers:
All the GUI tests in travis failed. That seems to be because of the older fife version available to travis. What's interesting is, that it does not fail with "we need a newer version of fife", which is bad.

LinuxDonald:
Here are all runetime errors in the code:
https://github.com/unknown-horizons/unknown-horizons/search?p=2&q=RuntimeError&utf8=%E2%9C%93

And here are the new typs that must be used:
https://github.com/fifengine/fifengine/commit/4c34e521d4160a9b5ea7890586404bdcacd9b310

MasterofJOKers:
I can do a grep, but thanks for the link to the exception types.
Still, I wanted to know, why he didn't convert the other few. My guess would be: because we don't know what exact error to fetch. @Niektory Is there a base class for all exceptions fife raises? Is fife.Exception that class? Why couldn't we use that one instead of the workaround, where we are not sure?

MasterofJOKers:
Oh right, the git version of fife we are using already says its 0.4.0, so that's why the tests don't fail earlier.

Niektory:
I went for minimal changes to avoid breaking anything. But maybe I was too conservative. I'll try updating all of them, then.
Yes, fife.Exception is the base class for all FIFE exceptions.

Niektory:
OK, done. I think this should do the trick.
I'm not sure about e137a4b. I assumed that code was supposed to handle the case of using an unknown font, which throws an `InitializationError` from Pychan (which isn't one of the new exception types so it seems it was already wrong before the recent FIFE change). So don't pull that commit if you believe I'm mistaken.

CrafterSvK:
Main branch doesn't work for me. This one does. thanks :)

LinuxDonald:
We have an other problem there too see line 114: http://pastebin.com/iPc2DEwK

Niektory:
@LinuxDonald That seems to be caused by #2516. I'll comment there.

PR #2586, opened 2016-12-25
Changed size of the text that was cut off

CrafterSvK:
Changed size from 225,80 to 225,70

screenshot_2016-12-25_17-08-23


CrafterSvK:
Still doesn't work... Problem occurs only with tools

CrafterSvK:
Problem fix. Please review it and merge it.

LinuxDonald:
Thx for the fix :)

PR #2590, opened 2017-01-04
E713 test for membership should be 'not in'

MarkusHackspacher:
E713 test for membership should be 'not in'


squiddy:
Please update `setup.cfg` to enable the check for E713. (https://github.com/unknown-horizons/unknown-horizons/blob/master/setup.cfg)

PR #2591, opened 2017-01-04
E713 test for membership should be 'not in'

MarkusHackspacher:


LinuxDonald:
@MasterofJOKers and @squiddy could you please review it?

horizons/gui/tabs/tabwidget.py
```diff
@@ -97,7 +97,7 @@ def show_tab(self, number):
 		"""Used as callback function for the TabButtons.
 		@param number: tab number that is to be shown.
 		"""
-		if not number in range(len(self._tabs)):
+		if number not in range(len(self._tabs)):
```
ChrisOelmueller: That check could probably be rewritten ;-)

MasterofJOKers: Yeah, we should use `xrange()` :D

PR #2594, opened 2017-01-06

MarkusHackspacher:
I make a PR with four .format() changes. I think that is not too big to check.


MasterofJOKers:
cherry-picked.

PR #2595, opened 2017-01-09
Add missing gettext imports to run_uh.py

Niektory:
Refs: #2516

### DIFF
--- a/run_uh.py
+++ b/run_uh.py
@@ -118,16 +118,17 @@ def excepthook_creator(outfilename):
 	to a file.
 	@param outfilename: a filename to append traceback to"""
 	def excepthook(exception_type, value, tb):
+		from horizons.i18n import gettext as T
 		with open(outfilename, 'a') as f:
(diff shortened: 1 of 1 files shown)


run_uh.py
```diff
@@ -176,6 +177,7 @@ def main():
 	try:
 		import yaml
 	except ImportError:
+		from horizons.i18n import gettext as _
```
MasterofJOKers: These lines don't seem to need the import, but rather need a removal of the `_()`, as other languages are only initialized on line 190 of this file in `horizons.main.start()` (which calls `horizons.engine.engine.Fife.init()`, which calls `horizons.engine.settings.Settings.apply()`, which calls `horizons.i18n.change_language()`).

Niektory:
Amended the PR.

Niektory:
Oh, right. That's what I get for only looking at old commits and not current code. Changed.

LinuxDonald:
Thank you very much for your help Niektory :) 👍

PR #2601, opened 2017-01-21
Porting UH to fife pychan

LinuxDonald:
Need to be done:

- [ ] Production Building gui fix
- [ ] Buildmenu sytle fixes
- [x] ESC ingame menu style fixes


LinuxDonald:
Needs for this branch:

https://github.com/fifengine/fifechan/tree/pychan_port
https://github.com/fifengine/fifengine/tree/pychan_porting

How to get the crash:
start new game

MasterofJOKers:
Starting a new game already crashes that? But it's the "configure_route.xml" that gets loaded, so I would have thought, some clicks are needed.

Anyway, the problem is, that the position.height or position.width returns 0. I don't know what changed in the fife(chan) branches, maybe @helios2000 could shine a little light on why there might be no width/height on an Icon any more.

LinuxDonald:
Okay the crash is now fixed. Lets tackle the gui. It needs alot of fixes.

squiddy:
What's the "fife pychan" branch about anyway? What's the motivation for these changes, I mean, why are we working on this?

helios2000:
Well, the idea behind the branch is to move most of Pychans functionality to Fifechan. Pychan should be not more as a wrapper for Fifechan. I've tried to make layouting as similar as possible but a few things changed to make it more consistent. The branch also includes a bunch of fixes and new features.
Here is the related milestone from fife: https://github.com/fifengine/fifengine/milestone/25

AndyMender:
The GUI still seems to freeze and occasionally crush the game. It usually happens when hovering over toolbar icons or selecting buildings. Since on UNIX systems run_uh.py suppresses both stdout and stderr via redirection to /dev/null, I'll have to remove the related lines to investigate further.

Meanwhile, a traceback related to a possible XML bug. The parser class method from pychan's attr.py receives a set of numbers that don't meet the expected format "x,y" or similar. The traceback is here: http://pastebin.com/jHPa653r

LinuxDonald:
This code maybe should port to xml file?

https://github.com/unknown-horizons/unknown-horizons/blob/pychan/horizons/gui/widgets/resourceoverviewbar.py#L621

Line 621 up to 637

The xml file is: https://github.com/unknown-horizons/unknown-horizons/blob/pychan/content/gui/xml/ingame/hud/resource_overview_bar_stats.xml

@MasterofJOKers do you know maybe how the xml should look like?

AndyMender:
Lines 622 and 623 in */horizons/gui/widgets/resourceoverviewbar.py mention ./content/gui/xml/ingame/hud/resource_overview_bar_stats.xml. Is that it? Then probably the "for" loop in 621-638 should be swapped with a wrapper to that xml? Is that possible via fifechan or would one need a Python library like libxml2?

In any case, I'd love to help.

MasterofJOKers:
What I don't get yet, is why that has to be done in python instead of doing it in the XML ... basically, these added HBoxes need a name based on their image and then we can set the translated help text on them. Maybe the `box.stylize()` also has to be done in python, but the other stuff should go into the XML I guess.

LinuxDonald:
In the left side top corner you can see how it must be look.

horizons/gui/widgets/resourceoverviewbar.py
```diff
@@ -624,13 +624,10 @@ def _init_stats_gui(self):
 			box = HBox(padding=0, min_size=(70, 0))
 			box.name = "resbar_stats_line_%s" % num
 			box.helptext = helptext
-			box.addChild(Icon(image=image, fixed_size=(16,16)))
+			box.addChild(Icon(image=image))
```
LinuxDonald: @MasterofJOKers could you please review the changes here?

LinuxDonald:
Started to rebranch it.

PR #2607, opened 2017-01-29
Change Python version interface to sys.version.

AndyMender:
sys.version is a string and by reading the first 3 characters we can get the major and minor Python release version more flexibly. This can be later further adjusted to fit Python3 if needed.

### DIFF
--- a/run_uh.py
+++ b/run_uh.py
@@ -67,10 +67,9 @@ def exit_with_error(title, message):
 
 def check_python_version():
 	# python up to version 2.6.1 returns an int. http://bugs.python.org/issue5561
-	if platform.python_version_tuple()[0] not in (2,'2'):
-		exit_with_error('Unsupported Python version', 'Python 2 is required to run Unknown Horizons.')
-
-
+    if float(sys.version[:3]) != 2.7:
+        exit_with_error('Unsupported Python version', 'Python 2.7 is required to run Unknown Horizons.')
+	
 check_python_version()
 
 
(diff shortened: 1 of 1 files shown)


run_uh.py
```diff
@@ -67,10 +67,9 @@ def exit_with_error(title, message):
 
 def check_python_version():
 	# python up to version 2.6.1 returns an int. http://bugs.python.org/issue5561
-	if platform.python_version_tuple()[0] not in (2,'2'):
-		exit_with_error('Unsupported Python version', 'Python 2 is required to run Unknown Horizons.')
-
-
+    if float(sys.version[:3]) < 2.7:
```
squiddy: That's certainly an interesting way to do it, but I don't think using `float` here is a good idea. Why not do this instead:

`if sys.version[:3] < (2, 7):`

The comment above can then be deleted, since it refers to `python_version_tuple` specifically.

AndyMender: I wasn't aware that sys.version_info can be sliced into a tuple. I agree that a tuple-to-tuple comparison is cheaper on memory, though I would write this instead:
`if sys.version_info[:2] != (2, 7):`
For now we assume Python 2.7 is the only fully supported version, and leaving it at "<" would not exclude Python 3.x which we don't support just yet :).

squiddy: Fine with me.

MasterofJOKers:
Merged with the following changes:
 * squashed into a single commit
 * Added one more newline between function and function call
 * Replaced spaces with tabs ( @AndyMender: UH uses tabs. Please take care of your editor configuration for future changes)

Other than that: :+1:

PR #2610, opened 2017-01-29
Some minor fixes to run_uh.py

AndyMender:
I briefly looked through run_uh.py and noticed a couple of things that could be cleaned up to make the code base easier to handle. After the final commit, the exit_with_error function can be simplified as no call in run_uh.py will actually satisfy: `title + "." == message`.

I know those changes aren't significant, but they will hopefully help in finding more severe issues in the long run :).

### DIFF
--- a/run_uh.py
+++ b/run_uh.py
@@ -128,7 +128,7 @@ def excepthook_creator(outfilename):
 		print(T('Please give it to us via IRC or our forum, for both see http://unknown-horizons.org .'))
 	return excepthook
 
-def exithandler(exitcode):
+def exithandler(exitcode, signum, frame):
 	"""Handles a kill quietly"""
 	signal.signal(signal.SIGINT, signal.SIG_IGN)
 	signal.signal(signal.SIGTERM, signal.SIG_IGN)
--- a/run_uh.py
+++ b/run_uh.py
@@ -121,9 +121,7 @@ def excepthook_creator(outfilename):
 		with open(outfilename, 'a') as f:
 			traceback.print_exception(exception_type, value, tb, file=f)
 		traceback.print_exception(exception_type, value, tb)
-		print('')
-		print(T('Unknown Horizons has crashed.'))
-		print('')
+		print('\n', T('Unknown Horizons has crashed.', '\n'))
 		print(T('We are very sorry for this and want to fix the underlying error.'))
 		print(T('In order to do this, we need the information from the logfile:'))
 		print(outfilename)
@@ -134,9 +132,7 @@ def exithandler(exitcode, signum, frame):
 	"""Handles a kill quietly"""
 	signal.signal(signal.SIGINT, signal.SIG_IGN)
 	signal.signal(signal.SIGTERM, signal.SIG_IGN)
-	print('')
-	print('Oh my god! They killed UH.')
-	print('You bastards!')
+	print('\nOh my god! They killed UH. \nYou bastards!')
 	if logfile:
 		logfile.close()
 	sys.exit(exitcode)
--- a/run_uh.py
+++ b/run_uh.py
@@ -93,7 +93,7 @@ def get_content_dir_parent_path():
 	options.append(os.path.dirname(os.path.realpath(unicode(__file__))))
 	# Try path for Mac Os X app container (Unknown Horizons.app).
 	# Unknown Horizons.app/Contents/Resources/contents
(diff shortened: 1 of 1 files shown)


run_uh.py
```diff
@@ -121,22 +121,18 @@ def excepthook(exception_type, value, tb):
 		with open(outfilename, 'a') as f:
 			traceback.print_exception(exception_type, value, tb, file=f)
 		traceback.print_exception(exception_type, value, tb)
-		print('')
-		print(T('Unknown Horizons has crashed.'))
-		print('')
+		print('\n', T('Unknown Horizons has crashed.', '\n'))
```
squiddy: I suppose that's a matter of preference, in the current code it's quite obvious where new lines are and how this will look like in the terminal.

AndyMender: I would always favor single print() calls, though I agree that here it doesn't matter so much. You can kick this commit out if you think it's redundant :).

run_uh.py
```diff
@@ -121,22 +121,18 @@ def excepthook(exception_type, value, tb):
...
 		print(T('We are very sorry for this and want to fix the underlying error.'))
 		print(T('In order to do this, we need the information from the logfile:'))
 		print(outfilename)
 		print(T('Please give it to us via IRC or our forum, for both see http://unknown-horizons.org .'))
 	return excepthook
 
-def exithandler(exitcode, signum, frame):
+def exithandler(exitcode):
```
squiddy: This change is wrong, `exithandler` needs to accept three arguments (and will be passed three arguments).

The `signal.signal` function expects a handler function that accepts two arguments: signum and frame. See here: https://docs.python.org/2/library/signal.html#signal.signal

It might be a bit confusing though, I can understand that. The `exithandler` function isn't called by us, but by the `signal` module.

```python
signal.signal(signal.SIGINT, functools.partial(exithandler, 130))
```

What `functools.partial(exithandler, 130)` does is return a new function that will pass through all arguments to `exithandler`, but pass 130 as first argument additionally.

AndyMender: I wasn't aware of that. Thank you for explaining the matter thoroughly.

run_uh.py
squiddy: Either way is probably fine.

However, talking about newlines beeing obvious, your change introduced an error.

`print('\n', T('Unknown Horizons has crashed.', '\n'))`

should be

`print('\n', T('Unknown Horizons has crashed.'), '\n')`

Notice the placement of the closing paranthesis.

run_uh.py
```diff
@@ -343,7 +339,7 @@ def setup_fife():
 	log_sys_info()
 	if not find_fife():
 		#TODO useful error message anyone?
```
squiddy: I'm wondering, can we drop this TODO then? Or what makes a useful error message here? Perhaps the list of locations we searched for?

LinuxDonald:
@AndyMender are you still working on this?

AndyMender:
I was very busy recently, my apologies!
Some of the changes I submitted were further adjusted by squiddy and if there are no conflicts, can probably be merged. 

Not sure about the TODO section. It seems run_uh.py looks for the fife module in a lot of places, while the default location is in most cases the library directory of a specific Python distribution (in our case, the main Python 2.7). Should I add some code so that _all_ of the searched locations are stored in a list and printed via a for-loop? Can someone comment on whether this is the right way to go?

LinuxDonald:
Maybe this is an good idea. Then they now where uh searches for fife.

squiddy:
@AndyMender I haven't adjusted any of your changes, I merely reviewed them. So far, there are still two issues (print statement, signal handler) that need to be resolved. If you could fix those, we can merge this.

If you want to work on the TODO section, go ahead, it's not not a blocker for a merge though.

AndyMender:
I undid/fixed the previous changes that accidentally introduced bugs and I'm now trying to make find_fife() and import_fife() a bit more useful. As in, to combine them in a way that the user gets a full list of where Python tried looking for the fife module.

squiddy:
> EDIT: I would like to consider the TODO section in a separate PR as it's a bit bigger and more involving, and I don't want to mix things.

Sounds good to me.

LinuxDonald:
@squiddy can that merged now into master?

squiddy:
@AndyMender thanks for your contribution. Feel free to tackle other issues. :)

PR #2611, opened 2017-01-31

MarkusHackspacher:
@MasterofJOKers could you review?

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -97,7 +97,7 @@ class ReRunInfoPlugin(Plugin):
 
 		_, module, call = test.address()
 
-		output = ['python2', 'run_tests.py', u'{0}:{1}'.format(module, call)]
+		output = ['python2', 'run_tests.py', u'{}:{}'.format(module, call)]
 
 		# add necessary flags
 		if 'tests.gui' in module:
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -97,7 +97,7 @@ class ReRunInfoPlugin(Plugin):
 
 		_, module, call = test.address()
 
-		output = ['python2', 'run_tests.py', u'%s:%s' % (module, call)]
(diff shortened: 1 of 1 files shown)


tests/utils.py
```diff
@@ -97,7 +97,7 @@ def formatError(self, test, err):
 
 		_, module, call = test.address()
 
-		output = ['python2', 'run_tests.py', u'%s:%s' % (module, call)]
+		output = ['python2', 'run_tests.py', u'{0}:{1}'.format(module, call)]
```
MasterofJOKers: weren't we going to use `{}` instead of `{0}`? You did use `{}` in you last Pull-Request.

MarkusHackspacher: With Python 2.6 is working `'{}'.format('a')`, but not `'{}:{}'.format('a', 'b')`. You have to use `'{0}:{1}'.format('a', 'b')`. But better shoud be `'{module}:{call}'.format(module='a', call='b')`.

MasterofJOKers: We don't support Python2.6 any more.

MasterofJOKers: And I don't think we need `{module}:{call}`, as it's only testing code that will not get translated and it's pretty obvious, that this is just a `':'.join([module, call])`.

squiddy: Yeah, `"{module}:{call}".format(module=module, call=call)` is definitly too much. And also, that's a good way to convince me to keep `"%s:%s" % (module, call)`. :D

But `"{}:{}".format(module, call)` is fine I guess.

tests/utils.py
```diff
@@ -97,7 +97,7 @@ def formatError(self, test, err):
 
 		_, module, call = test.address()
 
-		output = ['python2', 'run_tests.py', u'%s:%s' % (module, call)]
+		output = ['python2', 'run_tests.py', u'{}:{}'.format(module, call)]
```
MarkusHackspacher: Done

PR #2614, opened 2017-02-19
Major adjustments to import_fife, find_fife and setup_fife

AndyMender:
While trying to work out the TODO point in the `setup_fife()` function definition, I noticed that `import_fife()` and `find_fife()` require major adjustments. The main rationale was that `import_fife()` and `find_fife()` both return Boolean values and duplicate fife import calls. Below a brief rundown of introduced changes:

- `import_fife()` was renamed to `find_fife()`. The previous name was confusing, because the function originally only checked whether fife can be found in the provided 'paths'. Also, this is the only function that really checks whether fife exists in the specified paths. Alternatively, one could add an earlier return value when `find_fife()` finds the fife module. This can still be evaluated in if-else statements, as return None == False is true.

- `find_fife()` was renamed to `get_fife_paths()` as that's what it does anyway. Now it returns a list of paths to be later checked by find_fife(). This makes for modular and fine-grained `find_fife(get_fife_paths())` checks. Extra paths can be easily added to `get_fife_paths()` if needed.

- `setup_fife()` was changed so that the check for fife importing is local and failure is handled via try-except with a call to `exit_with_error()` if importing fails. Also, I added a pretty-print for the list of checked paths :).

- some other fixes to make the code easier to read.

I found it easier to introduce all changes as a single commit since anyhow those functions work in unison. If anything else uses them and now it's borked, I'll try to introduce fixes where needed.

Finally, now the code in `exit_with_error()` can be simplified so that the 'title' and 'message' arguments are not evaluated. That would be a separate PR, though :).

### DIFF
--- a/run_uh.py
+++ b/run_uh.py
@@ -282,7 +282,7 @@ def import_fife(paths):
 		try:
 			from fife import fife
 		except ImportError as e:
-			if str(e) != 'cannot import name fife':
+			if str(e) != "cannot import name 'fife'":
 				log().warning('Failed to use FIFE from %s', fife)
 				log().warning(str(e))
 				if str(e) == 'DLL load failed: %1 is not a valid Win32 application.':
--- a/run_uh.py
+++ b/run_uh.py
@@ -274,11 +274,12 @@ def setup_debugging(options):
 
 		log_sys_info()
 
-def import_fife(paths):
+def find_fife(paths):
+    """Returns True if the fife module was found in one of the supplied paths."""
 	try:
-		# If FIFE can't be found then this call will throw an exception.
-		settings = imp.find_module('fife', paths)
-		fife = imp.load_module('fife', *settings)
+        # If FIFE can't be found then this call will throw an ImportError exception:
+		module_info = imp.find_module('fife', paths)
+		fife = imp.load_module('fife', *module_info)
 		try:
 			from fife import fife
 		except ImportError as e:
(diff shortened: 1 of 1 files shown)


run_uh.py
```diff
@@ -294,53 +295,55 @@ def import_fife(paths):
...
 	return True
 
-def find_fife():
-	# Use the path the user provided.
+def get_fife_paths():
+	"""Returns all possible paths to search for the fife module. New paths to be added as needed."""
+	# Use the path the user provided:
 	from horizons.util.cmdlineoptions import get_option_parser
 	options = get_option_parser().parse_args()[0]
+	# All paths collected in a list:
```
squiddy: I don't think this comment improves the code, `paths = []` is already pretty clear. I'd remove it.

run_uh.py
```diff
@@ -294,53 +295,55 @@ def import_fife(paths):
...
+	paths = []
+	# Check if path was provided by user:
 	if options.fife_path:
 		fife_path = os.path.abspath(options.fife_path)
+		paths.append(fife_path)
 		# Support giving the path to FIFE_ROOT/engine/python/fife/__init__.pyc etc.
+
```
squiddy: That newline is confusing, the comment is for the code below, so it should be removed.

run_uh.py
```diff
@@ -294,53 +295,55 @@ def import_fife(paths):
...
-		exit(1)
-
-	# Try to use the default FIFE (equivalent of just trying to import it).
-	if import_fife(None):
-		return True
-
-	# Look for FIFE in the neighborhood of the game dir.
-	paths = []
+			paths.append(fife_path)
+			# Support giving the path to FIFE_ROOT/engine/python
```
squiddy: The comments should go before the line they belong to, not after.

run_uh.py
```diff
@@ -294,53 +295,55 @@ def import_fife(paths):
...
+	# Look for FIFE in the neighborhood of the game dir:
 	for opt1 in ('.', '..', '..' + os.sep + '..'):
 		for opt2 in ('.', 'fife', 'FIFE', 'Fife', 'fifengine'):
 			for opt3 in ('.', 'trunk'):
 				path = os.path.abspath(os.path.join('.', opt1, opt2, opt3, 'engine', 'python'))
 				if os.path.exists(path):
 					paths.append(path)
-	return import_fife(paths)
+
+	# Return collected list of fife search paths:
```
squiddy: Comment can be dropped as well as it doesn't add anything.

run_uh.py
```diff
@@ -294,53 +295,55 @@ def import_fife(paths):
...
-		exit_with_error('Failed to find and/or load FIFE', 'Make sure FIFE was installed as a package/redistributable or compiled from source.')
+	paths = get_fife_paths()
+	if not find_fife(paths):
+		try:
+			from fife import fife
+		except ImportError:
+			exit_string = ""
+			for path in paths:
+				exit_string += "{}\n"
+			exit_with_error('Failed to find and/or load FIFE', 'Below directory paths were tested:\n' + exit_string.format(*paths))
```
squiddy: Putting together that text with the directories can be simplified like this:

```python
directories = '\n'.join(paths)
exit_with_error('Failed to find and/or load FIFE', 'Below directory paths were tested:\n' + directories)
```

run_uh.py
```diff
@@ -294,53 +295,55 @@ def import_fife(paths):
...
-def find_fife():
-	# Use the path the user provided.
+def get_fife_paths():
+	"""Returns all possible paths to search for the fife module. New paths to be added as needed."""
+	# Use the path the user provided:
 	from horizons.util.cmdlineoptions import get_option_parser
 	options = get_option_parser().parse_args()[0]
+	# All paths collected in a list:
+	paths = []
+	# Check if path was provided by user:
```
squiddy: Please keep the comments ending in an optional dot, not a colon.

squiddy:
Haven't tested it yet, but it's looking like a good change.

run_uh.py
AndyMender: This was confusing to me in the original code as newlines were extremely sparse. I myself always try to leave space between even the smallest blocks to make the code readable. I'll try to re-arrange this bit properly, though.

run_uh.py
AndyMender: Yes, your change is better :). Sorry for trying to make it more complicated than necessary.

run_uh.py
```diff
@@ -294,53 +295,50 @@ def import_fife(paths):
...
-		exit_with_error('Failed to find and/or load FIFE', 'Make sure FIFE was installed as a package/redistributable or compiled from source.')
+	paths = get_fife_paths()
+	if not find_fife(paths):
+		try:
+			from fife import fife
+		except ImportError:
+			directories = '\n'.join(paths)
+			exit_with_error('Failed to load module fife', 'Below directory paths were tested:\n' + directories)
 
+	# Another import call still needed?
```
squiddy: I think it is needed. We do import fife before (line 336), but that code isn't always executed. The only other import happens in `find_fife` which isn't run in the global scope.

PR #2615, opened 2017-02-21
Fix layout for hotkeys menu

citizenjules:
Fixes broken label positions in hotkeys menu.


LinuxDonald:
Thx for this patch nice work :)

PR #2616, opened 2017-02-23
Run uh fixes2

AndyMender:
Very minor fixes to the `exit_with_error()` function. Since the TODO later in the code was addressed, the `title + '.' != message` check is not needed anymore. In fact, it will never return True. At least within run_uh.py.

Next, I will help LinuxDonald with pychan :).

### DIFF
--- a/run_uh.py
+++ b/run_uh.py
@@ -50,7 +50,7 @@ import traceback
 # this will break any run_uh imports from other locations (e.g. _get_version())
 
 def exit_with_error(title, message):
-	print(title + '\n' + message);
+	print('Error: ' + title + '\n' + message);
 	try:
 		import Tkinter
 		import tkMessageBox
@@ -58,7 +58,7 @@ def exit_with_error(title, message):
 		window.wm_withdraw()
 		tkMessageBox.showerror(title, message)
 	except ImportError:
-		print("Tkinter module missing. Install Tkinter for Python2.7 to view pop-up error dialogs");
+		pass
 	exit(1)
 
 def check_python_version():
--- a/run_uh.py
+++ b/run_uh.py
@@ -50,6 +50,7 @@ import traceback
 # this will break any run_uh imports from other locations (e.g. _get_version())
 
 def exit_with_error(title, message):
+	print(title + '\n' + message);
 	try:
 		import Tkinter
 		import tkMessageBox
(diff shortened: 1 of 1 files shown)


LinuxDonald:
@AndyMender i have corrected my Nick ;) i hope @squiddy will review it.

squiddy:
I'll have a look later today.

run_uh.py
```diff
@@ -50,19 +50,15 @@
...
 	try:
 		import Tkinter
 		import tkMessageBox
 		window = Tkinter.Tk()
 		window.wm_withdraw()
 		tkMessageBox.showerror(title, message)
 	except ImportError:
-		# tkinter may be missing
-		pass
+		print("Tkinter module missing. Install Tkinter for Python2.7 to view pop-up error dialogs");
```
squiddy: I'd like to see this visually separated from the actual error message. Since the module is optional, it shouldn't shadow the actual problem. Right now, it'll probably look like this:

```
Unsupported Python version
32 bit FIFE requires 32 bit (x86) Python 2.
Tkinter module missing. Install Tkinter for Python2.7 to view pop-up error dialogs
```

Looking at this, title and message should stand out more. Perhaps like this:

```
Error: Unsupported Python version
32 bit FIFE requires 32 bit (x86) Python 2.
Tkinter module missing. Install Tkinter for Python2.7 to view pop-up error dialogs
```

I suppose it's enough to add a new empty line:

```
Error: Unsupported Python version
32 bit FIFE requires 32 bit (x86) Python 2.

Tkinter module missing. Install Tkinter for Python2.7 to view pop-up error dialogs
```

I was going to suggest rewording the sentence, but then again.. this function is used a total of three times. Two times to show an error about python and one time about fife. Is it even *necessary* to ask the user to install tkinter?

squiddy: But I leave that up to you, but I'd like to see at least a newline between error and tkinter notice.

AndyMender: Yes, I agree that the need for an extra debugging message in `exit_with_error()` is somewhat debatable. As alternatives I see:
- placing a try-except at the beginning of the run_uh.py code, possibly in the section with all of the other imports
- add a note in the documentation ("If you want to see additional error pop-ups, make sure your Python2.7 was installed with Tkinter") -> **recommended as it's the simplest solution**

The thing is that Tkinter comes with Python on many, but not all platforms. For instance, on FreeBSD it's an optional module with all of the Python packages, but the base Python packages do not contain it. On the other hand, missing Tkinter doesn't break anything as the error message printed by `exit_with_error()` is sent to stdout anyhow. 

I will therefore remove it and restore the `pass` instruction.

PR #2617, opened 2017-02-24

MarkusHackspacher:
change to the `.format()` function in this three files:
tests/gui/logger.py
tests/gui/test_editor.py
tests/gui/test_editor.py
@MasterofJOKers could you review?

### DIFF
--- a/tests/gui/logger.py
+++ b/tests/gui/logger.py
@@ -266,7 +266,7 @@ class TestCodeGenerator(object):
 			if self._last_slider_event:
 				w = self._last_slider_event
 				if w.name != widget.name:
-					self._add(["gui.find('{}').slide({})".format(w.name, w.value), ""])
+					self._add(["gui.find('{}').slide({:f})".format(w.name, w.value), ""])
 					self._last_slider_event = None
 
 			if isinstance(widget, widgets.ListBox):
--- a/tests/gui/logger.py
+++ b/tests/gui/logger.py
@@ -338,9 +338,11 @@ class TestCodeGenerator(object):
 
 		# Output debug information, no test code yet
 		if button:
-			log.debug("# {}.{}({}, {}, '{}')".format(tool_name, event_name, x, y, button))
+			log.debug("# {}.{}({:d}, {:d}, '{}')".format(
+				tool_name, event_name, x, y, button))
 		else:
-			log.debug("# {}.{}({}, {})".format(tool_name, event_name, x, y))
+			log.debug("# {}.{}({:d}, {:d})".format(
+				tool_name, event_name, x, y))
 
 	def dialog_opened(self):
 		"""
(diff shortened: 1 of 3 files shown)


tests/gui/logger.py
```diff
@@ -266,24 +266,26 @@ def new_widget_event(self, widget, event_name, group_name):
 			if self._last_slider_event:
 				w = self._last_slider_event
 				if w.name != widget.name:
-					self._add(["gui.find('%s').slide(%f)" % (w.name, w.value), ""])
+					self._add(["gui.find('{}').slide({})".format(w.name, w.value), ""])
```
MasterofJOKers: You've lost the `f` here.

tests/gui/logger.py
```diff
@@ -332,13 +334,13 @@ def new_mousetool_event(self, tool_name, event_name, x, y, button):
...
-			raise Exception("Event '%s' not supported." % event_name)
+			raise Exception("Event '{}' not supported.".format(event_name))
 
 		# Output debug information, no test code yet
 		if button:
-			log.debug("# %s.%s(%d, %d, '%s')" % (tool_name, event_name, x, y, button))
+			log.debug("# {}.{}({}, {}, '{}')".format(tool_name, event_name, x, y, button))
 		else:
-			log.debug("# %s.%s(%d, %d)" % (tool_name, event_name, x, y))
+			log.debug("# {}.{}({}, {})".format(tool_name, event_name, x, y))
```
MasterofJOKers: You've lost the `d`s here.

MarkusHackspacher: Other question this is not to easy to read:
`log.debug("# {}.{}({}, {})".format(tool_name, event_name, x, y))`
What do you think, is this maybe better?
`log.debug("# {tool}.{event}({:d}, {:d})".format(tool=tool_name, event=event_name, x, y))`

MasterofJOKers: I don't find `'# {}.{}({:d}, {:d})'` hard to read.

MarkusHackspacher:
Sorry for the lot of commits, but I oversee a `d`.

AndyMender:
Markus, you can send all of the changes as a single commit :). If a Travis build fails, it means something is not OK. Look at the log to see what went wrong.

I agree with MasterofJOKers, though. Dropping the C-style modulo substitution operator requires to specify the data type in the .format substitution field { }. However, this has the added boon that the Python interpreter will more rigidly check the data type of the passed value. 

Sometimes the fact that Python defaults to a specific data type (most likely a string literal?) is problematic and in other cases it's mandatory to specify the expected data type (for instance, when generating fixed spacing for string literals). For the sake of correctness, it's better to simply always include it.

MarkusHackspacher:
See #2618

PR #2620, opened 2017-02-25

MarkusHackspacher:
Change to the .format() function in some file in the  horizons/world/ direction.

### DIFF
--- a/horizons/world/units/collectors/collector.py
+++ b/horizons/world/units/collectors/collector.py
@@ -396,8 +396,8 @@ class Collector(Unit):
 				new_reslist.append( entry )
 
 			remnant = self.get_component(StorageComponent).inventory.alter(entry.res, actual_amount)
-			assert remnant == 0, "{} couldn't take all of res {}; remnant: {}; planned: {amount}".format(
-			       self, entry.res, remnant, amount=entry.amount)
+			assert remnant == 0, "{} couldn't take all of res {}; remnant: {}; planned: {}".format(
+			       self, entry.res, remnant, entry.amount)
 		self.job.reslist = new_reslist
 
 	def transfer_res_to_home(self, res, amount):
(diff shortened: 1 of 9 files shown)


horizons/world/units/collectors/collector.py
```diff
@@ -396,8 +396,8 @@ def transfer_res_from_target(self):
 				new_reslist.append( entry )
 
 			remnant = self.get_component(StorageComponent).inventory.alter(entry.res, actual_amount)
-			assert remnant == 0, "%s couldn't take all of res %s; remnant: %s; planned: %s" % \
-			       (self, entry.res, remnant, entry.amount)
+			assert remnant == 0, "{} couldn't take all of res {}; remnant: {}; planned: {amount}".format(
```
squiddy: Why did you name make the last argument (`amount`) to format? Just use `{}` like the others.

horizons/world/units/collectors/collector.py
```diff
@@ -406,8 +406,8 @@ def transfer_res_to_home(self, res, amount):
 		remnant = self.get_home_inventory().alter(res, amount)
 		#assert remnant == 0, "Home building could not take all resources from collector."
 		remnant = self.get_component(StorageComponent).inventory.alter(res, -amount)
-		assert remnant == 0, "%s couldn't give all of res %s; remnant: %s; inventory: %s" % \
-		       (self, res, remnant, self.get_component(StorageComponent).inventory)
+		assert remnant == 0, "{} couldn't give all of res {}; remnant: {}; inventory: {inv}".format(
```
squiddy: Why did you name make the last argument (`inv`) to format? Just use `{}` like the others.

AndyMender:
What about using `{:s}` instead of `{}`? It's not necessary, because Python defaults to string literals when the substitution field is left empty? Similar cases for floats and decimals were raised [here](https://github.com/unknown-horizons/unknown-horizons/pull/2617).

squiddy:
> What about using {:s} instead of {}?

I think we've decided somewhere in an earlier PR that we just use `{}`.

> Similar cases for floats and decimals were raised here.

In these cases, the previous code used explicit `%d` format modifiers, and initially these were changed to `{}`, which changed the behaviour.

AndyMender:
I'll bear this in mind, thank you for the clarification! :)

PR #2621, opened 2017-02-26

MarkusHackspacher:
Change to the .format() function in some file in the horizons/world/ direction.

### DIFF
--- a/horizons/savegamemanager.py
+++ b/horizons/savegamemanager.py
@@ -204,10 +204,10 @@ class SavegameManager(object):
 				os.unlink(filename)
 
 		if autosaves:
-			tmp_del("%s/*.%s" % (cls.autosave_dir, cls.savegame_extension),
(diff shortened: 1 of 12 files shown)


horizons/savegamemanager.py
```diff
@@ -400,7 +400,7 @@ def get_scenario_metadata(cls, scenario="", filename=""):
 	def get_savegamename_from_filename(cls, savegamefile):
 		"""Returns a displayable name, extracted from a filename"""
 		name = os.path.basename(savegamefile)
-		name = name.rsplit(".%s"%cls.savegame_extension, 1)[0]
+		name = name.rsplit(".{}".format(cls.savegame_extension), 1)[0]
```
squiddy: Not related to your changes, but what kind of code is that?

I think it could be written as

```python
name = os.path.splitext(os.path.basename(savegamefile))[0]
cls.log.debug("Savegamemanager: savegamename: %s", name)
return name
```

PR #2625, opened 2017-02-28
Simplify get_savegamename_from_filename

MarkusHackspacher:
Is base of the idea of @squiddy 
https://github.com/unknown-horizons/unknown-horizons/pull/2621#discussion_r103399508


squiddy:
I'd really appreciate it if you could word the commit message a bit better. Something like `Simplify get_savegamename_from_filename` is so much more meaningful than `better code`.

MarkusHackspacher:
Or you could make a commit with this change, you have see and solve it.
I make just a remainder.

MarkusHackspacher:
#2626

PR #2627, opened 2017-02-28

MarkusHackspacher:
Change to the .format() function in some file in the horizons/network/ and / directory.

In the moment is Travis CI not working:
https://www.traviscistatus.com/incidents/hmwq9yy5dh9d

### DIFF
--- a/horizons/network/common.py
+++ b/horizons/network/common.py
@@ -145,7 +145,7 @@ class Game(object):
 
 		def __str__(self):
 			strvals = ["Open", "Prepare", "Running"]
-			return "{}".format(strvals[self.state])
+			return strvals[self.state]
 
 	def __init__(self, packet, creator):
 		# pickle doesn't use all of these attributes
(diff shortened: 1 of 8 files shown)


horizons/network/common.py
```diff
@@ -145,7 +145,7 @@ def __init__(self, state=Open):
 
 		def __str__(self):
 			strvals = ["Open", "Prepare", "Running"]
-			return "%s" % (strvals[self.state])
+			return "{}".format(strvals[self.state])
```
squiddy: Please change that to `return strvals[self.state]`.

It's not related to your changes, but it's a simple thing to do and we might forget about this later.

horizons/network/common.py
```diff
@@ -250,6 +251,6 @@ def __init__(self, state=NotSet):
 
 	def __str__(self):
 		strvals = ["NotSet", "TerminateGame"]
-		return "%s" % (strvals[self.state])
+		return "{}".format(strvals[self.state])
```
squiddy: See comment above.

horizons/network/common.py
squiddy: (dummy comment to set review to 'request changes')

PR #2628, opened 2017-03-01

MarkusHackspacher:
Change to the .format() function in some file in the development/ directory.


squiddy:
If there are other files in development that need fixing and you plan on fixing them, please do it in this pull request. 

It is a bit tiring to review this every day, I'd prefer bigger chunks.

MarkusHackspacher:
Change file in development. And solve merge-conflict in development/generate_credits.py

LinuxDonald:
@squiddy can this here be merged?

development/print_db_data.py
```diff
@@ -122,50 +122,36 @@ def print_production_lines():
 def print_verbose_lines():
 	print('Data has been moved, this view is unavailable for now')
 	return
-	def _output_helper_prodlines(string, list):
```
LinuxDonald: why it that part of code get delete?

MarkusHackspacher: When a function calls return , processing for that function stops and control is returned to the code that called the function. Any code that is placed after a return statement will never be executed. If any part of the program could potentially rely on this code, then the code must be moved to a location where it can be reached. If the code is unnecessary, then it should be removed, because extra code reduces readability.

LinuxDonald:
@squiddy can this get merged?

development/print_db_data.py
ChrisOelmueller: So that leaves the question: Is the code unnecessary? I understand you do not feel like fixing it, but maybe others would? Removing the old implementation entirely is going to make that much less likely.

Least you could do, in my opinion: open a ticket about this functionality (if there isn't one yet), include some sample output and a reference to the removed code there, and insert that ticket into the "has been moved" message line above.

MarkusHackspacher: Hello @ChrisOelmueller , first point, I had to open a ticket about this. That is right. My fault. second point, but I do not really dare to write, if I write a request, please to delete this file, the file is increased since 2013 only by the year number, alone the file name print_db_data.py, who please print a database, today in 2017?
https://github.com/unknown-horizons/unknown-horizons/commits/master/development/print_db_data.py

ChrisOelmueller: People who would like to obtain information about UH's data files in a condensed, easy-to-understand format. That should include all developers at the very least.

squiddy:
Merged in 241b2ac1fe020df8a2740e14928b29992682a41d, but restored the deleted code.

PR #2490, opened 2016-09-24
Python 3

squiddy:
Needs https://github.com/fifengine/fifengine/tree/python3.

untitled
# TODO
- [x] Make tests work again. Our fife import hack fails with metaclass conflicts.
- [ ] Think of a better way to mock fife for unit/integration tests

Fix failing tests
- [ ] tests.game.test_savegame_upgrade:test_upgrade
- [ ] tests.game.test_traderoute:test_traderoute_basic
- [x] tests.game.test_load_save:test_hunter_save_load
- [x] Improve parts of the code, e.g. `list(zip(...))[0]` should be replaceable with `next(zip(...))`
- [x] Fix nose stdout capture when running gui tests (fails with InvalidOperation: fileno)
- [x] `Warning: Failed to open /tmp/tmp8cv7hzdq/yamldata.cache as cache: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte`
- [x] `TypeError: 'float' object cannot be interpreted as an integer` shapes.circle:67. Probably / vs. //
- [ ] Lots of `Object 117269 of type 1 does not know when to pay its rent. Disregard this when loading old savegames or on running cost changes.` messages during tests
- [ ] Check uses of division operator

PR #2639, opened 2017-03-10
Py2 to py3 Work in Progress

LinuxDonald:
New dep for python 3 port:

python-future


LinuxDonald:
@squiddy should we merge his work into an new branch? that you maybe could help him?

squiddy:
Might as well work on/push to master. No one is going to review this mix of python 3 port and random code "improvements" / comments with questions.

AndyMender:
So I hit a major blocker on this one. Apparently, there is some endless recursion in the way the _LazyString class overloads the __getattr__ operator or something upstream of it generates an endless loop. Since this is the translation system, it's a bit out of my league.

The initial traceback is here: http://pastebin.com/4ev3xd2L

Notice that I needed to change the return value in text2gui some commits back as the unicodePolicy no longer applies (all strings are Unicode UTF-8 by default).  According to the traceback, that's directly upstream of the endless recursion. However, this merely simplifies the return value.

Niektory:
Possibly related? https://github.com/mitsuhiko/speaklater/pull/4

AndyMender:
Unfortunately, yes. I'm afraid that this is the very same problem. I applied the changes from the commit you linked, but that merely highlights the actual issue.

squiddy:
Did you have a look at the previous python 3 port? The solution might be there already.

AndyMender:
At one point I got slightly confused by the` gettext` imports, because the function is named exactly the same way, yet the gettext from horizons/i18n also wraps in the global `_trans`. My next commit fixes this [(.](https://github.com/unknown-horizons/unknown-horizons/pull/2639/commits/f98c0f19d03d9a09b270c19318caba10abf7c1a4)

AndyMender:
New issue in horizons/gui/util.py. Traceback available here: http://pastebin.com/5ejxVnmR.
Several things are happening and I tried adding 2 `print() `calls to see what 's wrong. Apparently, the content/gui/xml/mainmenu/loadingscreen.xml file is not loaded properly. 

@LinuxDonald, might you have an idea?

AndyMender:
Quick update. For now I managed to get the UH gui to launch (main menu), but there is still plenty of work to do :).

AndyMender:
For some reason the `Build `class in horizons/command/building.py returns `None`  when called in horizons/world/worldutils.py. I will have a look later why this is the case.

horizons/i18n/__init__.py
```diff
@@ -153,4 +151,4 @@ def change_language(language=None):
 	fontdef = get_fontdef_for_locale(new_locale)
 	horizons.globals.fife.pychan.loadFonts(fontdef)
 
-	LanguageChanged.broadcast(None)
+LanguageChanged.broadcast(None)
```
squiddy: I don't think this is correct.

squiddy:
> I need some help on this one.

I can only suggest to run the test suite. It probably gives you more hints about where things are going wrong.

AndyMender:
Basic refactoring is done in the sense that the current code is compatible with Python3, however:
- issues that were present in the code previously were not improved (things like `list(zip(tuple_a, tuple_b))[0]` that @ChrisOelmueller noticed)
- the game, though launches and some of the features work, is still extremely unstable and things like randomly appearing `None` return values need to be addressed

To this end, I will continue working on the issues at hand, but I think we would need to address each of them separately as their sources might be unrelated.

LinuxDonald:
@AndyMender is it ready for testing? Any know crashes? Ready for review?

AndyMender:
@LinuxDonald, the codebase is ready for review and testing. Be aware that this is so far only a direct refactoring to Python3 with minor fixes. I haven't tested everything yet, so things are bound to not work properly yet. Of the things I still have issues with:
- The Python3 interpreter occasionally crashes with errors in `free()` memory management. This is rare, but happens.
- `Point` class  x and y coordinates are floats for some reason, despite the fact that horizons/util/shapes/* scripts make sure that everything is either converted to int or asserted as int. I still need to find the cause of this.
- I merged some changes from my python3_cleanup branch, because the game would crash on FreeBSD instantly due to the default use of ascii and not utf-8 encoding for `open()`.

AndyMender:
As agreed upon with @LinuxDonald , imminent fixes, even minor, will go to this branch.

LinuxDonald:
Yeah this branch needs fixes before it can be reviewed. Because UH have crash bugs..

LinuxDonald:
@AndyMender https://pastebin.com/fTT99HnD (open load game on main menu)

AndyMender:
On it, but first need to address the issue of game saving. As it is now, it causes this: https://pastebin.com/gXE0ZCCn

I will start running unit tests from the "tests" directory also to make it more efficient.

squiddy:
> I will start running unit tests from the "tests" directory also to make it more efficient.

You should configure CI to actually run the tests with python 3: https://github.com/unknown-horizons/unknown-horizons/pull/2490/commits/93b07c49544b41ff203d4f10d0c7c7bd08d3e8a9

LinuxDonald:
@AndyMender see https://github.com/unknown-horizons/unknown-horizons/commit/a8c03613eacc9e4685c439da0626624c82d7ad44#diff-78581b6469426a65d797d5fd7892db98

LinuxDonald:
@AndyMender because of unicode error look at @squiddy his old py3 branch i think he had fixed this bug there

LinuxDonald:
@AndyMender https://github.com/AndyMender/unknown-horizons/pull/2

PR #2640, opened 2017-03-18
Tier5 [Work-in-Progress]

LinuxDonald:
I will split this pr into many small ones for faster and easier merge.


horizons/constants.py
```diff
@@ -500,7 +500,7 @@ class TIER:
 
 	LOWEST = SAILORS
 	HIGHEST = ARISTOCRATS
-	CURRENT_MAX = CITIZENS
+	CURRENT_MAX = MERCHANTS
```
AndyMender: After several rounds of trial-and-error, this change causes the `test_settler_level_save_load` test to fail. Will try to investigate further.

LinuxDonald:
See #2745

LinuxDonald:
@nihathrael @AndyMender reviwe would be nice and maybe some git log cleanup.....

AndyMender:
@nihathrael I think we can squash it and then see what gives. There is a lot of old commits in this. What do you think?

LinuxDonald:
@jmdejong at this pr you could take an look too? Maybe if you want you could do too an rebase.

horizons/constants.py
```diff
@@ -246,6 +246,8 @@ class BUILDINGS:
 	GOLD_DEPOSIT     = 91
 	GEM_DEPOSIT      = 92
 
+  AMBIENT          = 72
```
baka0815: Wrong Indentation.

AndyMender: Please, use the provided .editorconfig file together with your IDE :). The UH project uses tabs instead of spaces and a couple of other minor tweaks.

Issue #2602, opened 2017-01-21
There should be a --mute or --no-audio option in unknown horizons 


shirishag75:
As of 

```
[$] unknown-horizons --version                                                                                                      
Unknown Horizons 2017.1
```
there is way to mute audio via CLI. The --help shows the following - 

```
─[$] unknown-horizons --help                                                                                                         
Usage: unknown-horizons [options]

Options:
  --version             show program's version number and exit
  -h, --help            show this help message and exit
  -d, --debug           Enable debug output to stderr and a logfile.
  --fife-path=<path>    Specify the path to FIFE root directory.
  --restore-settings    Restores the default settings. Useful if Unknown
                        Horizons crashes on startup due to misconfiguration.
  --mp-master=<ip:port>
                        Specify alternative multiplayer master server.
  --mp-bind=<ip:port>   Specify network address to bind local network client
                        to. This is useful if NAT holepunching is not working
                        but you can forward a static port.

  Starting Unknown Horizons:
    --start-map=<map>   Starts <map>. <map> is the mapname.
    --start-random-map  Starts a random map.
    --start-specific-random-map=<seed>
                        Starts a random map with seed <seed>.
    --start-scenario=<scenario>
                        Starts <scenario>. <scenario> is the scenarioname.
    --start-dev-map     Starts the development map without displaying the main
                        menu.
    --load-game=<game>  Loads a saved game. <game> is the saved game's name.
    --load-last-quicksave
                        Loads the last quicksave.
    --edit-map=<map>    Edit map <map>.
    --edit-game-map=<game>
                        Edit the map from the saved game <game>.

  AI options:
    --ai-players=<ai_players>
                        Uses <ai_players> AI players (excludes the possible
                        human-AI hybrid; defaults to 0).
    --human-ai-hybrid   Makes the human player a human-AI hybrid (for
                        development only).
    --force-player-id=<force_player_id>
                        Set the player with id <force_player_id> as the active
                        (human) player.
    --ai-highlights     Shows AI plans as highlights (for development only).
    --ai-combat-highlights
                        Highlights combat ranges for units controlled by AI
                        Players (for development only).

  Development options:
    --debug-log-only    Write debug output only to logfile, not to console.
                        Implies -d.
    --debug-module=<module>
                        Enable logging for a certain logging module (for
                        developing only).
    --logfile=<filename>
                        Writes log to <filename> instead of to the uh-userdir
    --profile           Enable profiling (for developing only).
[...]

PR #2644, opened 2017-04-11
closes #2542

melloskitten:
I adjusted the conditions so that there are no trees placed on sand tiles.

PR #2647, opened 2017-04-18
No audio option added - issue #2602

melloskitten:
I added a new option as suggested in the issue.


horizons/main.py
```diff
@@ -154,6 +154,10 @@ def start(_command_line_arguments):
 	if not setup_gui_logger(command_line_arguments):
 		return False
 
+	## Check if the no-audio flag has been set.
```
squiddy: These should be single `#`, just like in line 161 and all the others.

AndyMender:
I fixed the comment as @squiddy suggested.

squiddy:
IIRC the previous contribution introduced these as well, so there are probably more places.

melloskitten:
@squiddy I'm sorry! It wasn't my intention. If it helps you guys, this is the other PR I had where I have three comments with ## instead of # https://github.com/unknown-horizons/unknown-horizons/pull/2644/files

LinuxDonald:
fixed.

squiddy:
> @squiddy I'm sorry! It wasn't my intention.

No problem, unfortunately I saw it too late. I don't think our code style checker is configured to warn about that either. :)

PR #2651, opened 2017-04-29
I#2507 - fix custom scenarios command-line option

AndyMender:
The `_find_matching_map` function required some rebuilding, which I did.

### DIFF
--- a/horizons/main.py
+++ b/horizons/main.py
@@ -445,7 +445,9 @@ def _load_cmd_map(savegame, ai_players, force_player_id=None):
 	return True
 
 def _find_matching_map(name_or_path, savegames):
-	"""*name_or_path* is either a map/savegame name or path to a map/savegame file."""
+	"""Find a map by name or path specifiec by user.
+	@param name_or_path: either a map/savegame name or path to a map/savegame file
+	@return: str with the path to the map/savegame file on success"""
 	game_language = horizons.globals.fife.get_locale()
 	# now we have "_en.yaml" which is set to language_extension variable
 	language_extension = '_' + game_language + '.' + SavegameManager.scenario_extension
--- a/horizons/main.py
+++ b/horizons/main.py
@@ -461,7 +461,7 @@ def _find_matching_map(name_or_path, savegames):
 			print("Error: '{name}' is not a valid Unknown Horizons map or savegame file.".format(name=name_or_path))
 			return
 
-	name = name_or_path	# name_or_path is a possible name of a savegame
+	name = name_or_path	# name_or_path is a name of a map/savegame
 
 	# Check if name matches any map/savegame name
 	if name not in savegames.keys():
--- a/horizons/main.py
+++ b/horizons/main.py
@@ -463,18 +463,15 @@ def _find_matching_map(name_or_path, savegames):
 	name = name_or_path	# name_or_path is a name of a map/savegame
 
 	# Check if name matches any map/savegame name
-	if name not in savegames.keys():
+	if name not in savegames:
 		print("Error: map/savegame {name} doesn't exist.".format(name=name))
 		return
 
 	# Check if name is ambiguous
-	found_names = ""
-	for test_name in savegames.keys():
-		if test_name.startswith(name):
-			found_names += name + "\n"
-	if len(found_names.splitlines()) > 1:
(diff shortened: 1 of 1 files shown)


squiddy:
Tests are failing, see isort output on Travis.

squiddy:
I'm going to take a look at the code this evening.

horizons/main.py
```diff
@@ -445,36 +445,45 @@ def _load_cmd_map(savegame, ai_players, force_player_id=None):
 	return True
 
 def _find_matching_map(name_or_path, savegames):
-	"""*name_or_path* is either a map/savegame name or path to a map/savegame file."""
+	"""Find a map by name or path specifiec by user.
+	@param name_or_path: either a map/savegame name or path to a map/savegame file
```
squiddy: If you introduce these kind of variable declarations here, might as well make them complete and add one for `savegames`.

horizons/main.py
```diff
@@ -445,36 +445,45 @@ def _load_cmd_map(savegame, ai_players, force_player_id=None):
...
-			print("Error: Found multiple matches:")
-			for name_or_path in map_file.splitlines():
-				print(os.path.basename(name_or_path))
-			return
-		else:
-			return map_file
-	else: # not a savegame, check for path to file or fail
-		if os.path.exists(name_or_path):
+
+	# Check if name_or_path is a path
```
squiddy: This comment is redundant and should be removed. `os.path.exists` doesn't need any explanation IMO.

horizons/main.py
```diff
@@ -445,36 +445,45 @@ def _load_cmd_map(savegame, ai_players, force_player_id=None):
...
 			return name_or_path
 		else:
-			print("Error: Cannot find savegame or map '{name}'.".format(name=name_or_path))
+			print("Error: '{name}' is not a valid Unknown Horizons map or savegame file.".format(name=name_or_path))
 			return
 
+	name = name_or_path	# name_or_path is a name of a map/savegame
+
+	# Check if name matches any map/savegame name
+	if name not in savegames.keys():
```
squiddy: The `keys` can be removed, this will automatically happen for dictionaries. `in` looks into elements for a list, or keys in dictionaries.

horizons/main.py
```diff
@@ -445,36 +445,45 @@ def _load_cmd_map(savegame, ai_players, force_player_id=None):
...
+		print("Error: map/savegame {name} doesn't exist.".format(name=name))
+		return
+
+	# Check if name is ambiguous
+	found_names = ""
```
squiddy: This whole block should be changed a bit. Building up a single string in case it might be shown later, and splitting it on newlines to check things is kinda backwards. Keep it in a list as long as possible and only convert it for display when you need it.

```python
found_names = [test_name for test_name in savegames if test_name.startswith(name)]
if len(found_names) > 1:
    print(...)
    print('\n'.join(found_names))
```

squiddy: Also, note how you can remove the `keys` here as well. Iteration and contains (`in`) in dictionaries work on the keys.

squiddy:
If possible, I'd like to see a (high-level) test here as well. Similar to this one: https://github.com/unknown-horizons/unknown-horizons/blob/master/tests/unittests/test_main.py

Basically, call horizons.main with the options that a call of `run_uh.py --start-scenario=tutorial` would generate, and assert that these options here (https://github.com/unknown-horizons/unknown-horizons/blob/master/horizons/main.py#L423) are correct.

This should help to prevent errors like these in the future. We probably have zero coverage of the command line interface.

horizons/main.py
AndyMender: True! I missed that, sorry :).

horizons/main.py
AndyMender: The workaround you pointed out was in the original code. I intended a more optimal solution altogether (with `list(savegames.keys()).count(name) > 1`) though that would then miss the output of duplicate entries.

I'll use your solution. Thanks!

AndyMender:
Unfortunately, I don't have much experience for writing unittests for something as complex as this project, so I might still need your help on that :).

squiddy:
@AndyMender I guess for now we can merge this and write a test later. I'll try to get some basics laid out so we can build upon it.

Issue #2219, opened 2013-12-04
Route menu is messy


acini:
![snapshot6](https://f.cloud.github.com/assets/6072881/1673224/9a9fba68-5ce3-11e3-86fe-64f6804ad691.png)
![snapshot12](https://f.cloud.github.com/assets/6072881/1673238/f2c56314-5ce3-11e3-8c64-cae831a53dac.png)
The route menu is not ergonomic.
1. The load/unload toggle button is confusing. It switches between 3 different states: 2 toggle states and 1 hover state. It could simply be 2 toggle states.
2. The red arrow can be placed before the ship icon and pointing away from it. This way, it will be give a solid feel making as ship icon stays in place.
3. The "scrollbar" which adjusts item load-unload amount should have easily clickable arrows on top and bottom ends. Maybe pressing shift with will could load 5x amount, with ctrl 10x etc.
4. Currently only the area taken by "X" or "Item" is clickable. The "18t" area and area on right side of resource picture can be made clickable so that player doesn't have to point to tiny pixel area everytime.
5. Large player names (here, Unnamed Traveler) overlaps load/unload slots. Maybe they could be made "Unnamed Tra..." etc. or reduce font size. Or maybe one could implement player's name character limit.
##

CrafterSvK:
Easiest thing is remove the hover thing. It is confusing and I had to learn how it works. That would be much easier :)

jorgesumle:
@CrafterSvK I like that feature :(

janexx:
I always don't know, wheter the hover message tells me the state how is it now or how it should be, when I click on an arrow. I am always confused with this menu.

LinuxDonald:
I think the gui needs that hover get removed and a little redesign.

CrafterSvK:
It is okay I guess but it is confusing. If you remove it. It iwll be much better

CrafterSvK:
#2467

LinuxDonald:
@CrafterSvK you could fix the other points from the issue too?

3. The "scrollbar" which adjusts item load-unload amount should have easily clickable arrows on top and bottom ends. Maybe pressing shift with will could load 5x amount, with ctrl 10x etc.
4. Currently only the area taken by "X" or "Item" is clickable. The "18t" area and area on right side of resource picture can be made clickable so that player doesn't have to point to tiny pixel area everytime.
[...]

PR #2652, opened 2017-04-30
Issue #2219 - adjust labels and positioning of trade slots.

LinuxDonald:


squiddy:
Why is this being merged when tests fail?

LinuxDonald:
Tests fails on master too before this merge. They dont have something todo with this pr.

squiddy:
Well, fix tests on master then? Our have you checked that this PR didn't introduce new issues?

LinuxDonald:
I dont have seen an new errors.

AndyMender:
`isort -c -rc horizons tests *.py` ran locally doesn't produce the errors that pop up on Travis CI.

LinuxDonald:
@squiddy i think it have something todo with the prebuild fife package from @MasterofJOKers not any code change on our side.

squiddy:
Ok. Can't check right now, but my guess is that the Fife package is put into a different import group (stdlib, third party, own) now and therefore isort notices seemingly wrong orders. The group can be overridden in isorts config though.

MasterofJOKers:
fife is configured in isort's config as "known_third_party": `known_third_party = fife`

squiddy:
So the issue is that `typing` is only included in Python 3.5 and newer. Which means in 3.4 it is a third-party module and therefore sorted differently. The reason why isort doesn't warn locally is that you simply haven't been developing against Python 3.4. If that's the version we're going to support, we should probably develop against that.

For now, I'm declaring typing as known standard library for isort so travis checks should be fixed then. Before packaging the next release, we need to make sure to include the dependency on typing.

PR #2654, opened 2017-05-01
Use up-to-date python libs on travis

LinuxDonald:
@squiddy on travis its green :) is it okay for merging into master?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,9 @@
-greenlet==0.4.10
-nose==1.3.7
-PyYAML==3.12
-Pillow==3.3.1
-isort==4.2.5
+greenlet
+nose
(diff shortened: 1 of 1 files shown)


squiddy:
I think I'd prefer the packages to be pinned to specific versions. Otherwise for example Travis might be failing randomly because it fetched a new version that isn't backwards compatible. 

We still can update the packages though while we're at it.

LinuxDonald:
Want we then pin they to the yet newest version?

squiddy:
Yes. Preferably a stable version.

squiddy:
I've thought about it, let's just remove the versions. You have to rebase onto master first to resolve conflicts.

PR #2662, opened 2017-05-03

MarkusHackspacher:

found because I look for E228 missing whitespace around modulo operator


squiddy:
Thanks.

PR #2664, opened 2017-05-06
remove change directory

MarkusHackspacher:
With this line can travis do the work without error only if it started from unknown-horizons/unknown-horizons.
Other user get an error back.


squiddy:
Ah, you mean if you setup Travis to test your fork.

Yeah, that line is obsolete since we're not building fife here anymore.

MarkusHackspacher:
Yes, I mean my travis test run, like this test: https://travis-ci.org/MarkusHackspacher/unknown-horizons/builds/228878424

PR #2666, opened 2017-05-06
use newest python libs on travis

LinuxDonald:


LinuxDonald:
@squiddy rebase done but create new because something gone wrong on rebase...

PR #2668, opened 2017-05-08
Move tests from nose to pytest

squiddy:
I really started to enjoy writing tests with pytest at work. The junit-style assertions and boilerplate of class based tests is sometimes a bit annoying, and pytests simple function-based tests are easy and quick to write. The approach to fixtures (you name specific features you want to use as parameters in test functions) may be a bit unusual, but I believe in the future we can reuse fixtures, perhaps even for gui tests, where you can define what you need.

`def test_something(dev_map, pirate_enabled)`

I believe pytest has more mature plugins (some I'm already using here, such as mocker and warnings), also right now, running tests in parallel seems to be working fine.

Things to do now:

- [x] Fix `tests.game.test_status_icons.test_settler_unhappy`
- [ ] Remove last dependencies to nose

```
session = <tests.game.SPTestSession object at 0x7fa8083c3400>
player = <horizons.world.player.HumanPlayer object at 0x7fa8084b7ef0>
    @game_test()
    def test_settler_unhappy(session, player):
    	settlement, island = settle(session)
    
    	cb = mock.Mock()
    	AddStatusIcon.subscribe(cb)
    
    	settler = Build(BUILDINGS.RESIDENTIAL, 30, 30, island, settlement=settlement)(player)
    
    	# certainly not unhappy
    	assert settler.happiness > 0.45
>   	assert not cb.called
E    AssertionError
```

- [ ] Get gui test `test_user_dir_contains_non_ascii` working.

```
    @pytest.mark.gui
    def test_user_dir_contains_non_ascii():
    	# NOTE we have to create the test this way because if it were defined globally,
    	# USER_DIR would not be defined yet at the time the decorator is evaluated
>   	gui_test(timeout=60, use_dev_map=True, _user_dir=USER_DIR)(dummy)()
E    TypeError: wrapped() missing 1 required positional argument: 'tmpdir'

tests/gui/test_unicode.py:55: TypeError
```

Things to do later:

- [ ] adjust more tests to make use of pytest features / convert them to function based tests
- [x] fix pytest assertions for unittests / game tests
- [ ] somehow enable pytest's assertions improvements for gui tests
- [ ] fix gui tests that rely on old savegame revisions
- [ ] somehow reimplement gui_test / game_test decorator to be easier to understand (not sure if this is easy to do with pytest)


squiddy:
And we don't need the `run_tests.py` wrapper anymore, that is nice!

AndyMender:
I got myself the `pytest` package as well. Great to see simplified interfaces instead of Java-like excessive class handling :).

AndyMender:
The tests/game/test_status_icons.py test fails at `assert not cb.called`. Is that intended behavior? Does  `AddStatusIcon.subscribe(cb)` not register it as "called"?

squiddy:
It is not intended, it's #1 on the to-do list here. And it's starting to annoy me. It works fine locally, I can trigger a failure only when I run tests in parallel.

I don't understand your second question. The issue is that cb shouldn't be called at all but it is. It's also random which one of the tests fail.

squiddy:
It only goes to show that even on tests docstrings are useful to explain what is being tested :/

AndyMender:
Apologies. I missed it! I think I should read a bit more on `pytest`.

AndyMender:
@squiddy, I tried running pytest from the Unknown Horizons root dir and I'm getting similar results as Travis. `tests.game.test_status_icons.test_settler_unhappy` doesn't seem to fail in my case:
https://pastebin.com/MQP1s8jS

squiddy:
Similar results? All tests are passing for you, same as for me. Ignore the
warnings and stuff, this is about a failing test. The rest is identical,
but the test run on Travis fails.

Am 11.05.2017 7:47 vorm. schrieb "Andrzej Bylicki" <notifications@github.com

squiddy:
> **EDIT: ** @squiddy, once I am suitably comfortable with pytest, I will help you write more unittests for Unknown Horizons :).

Glad to hear.

Although, I think just getting started writing tests is simpler. You don't really need all the functionality of pytest in the beginning. You should be somewhat familiar with `unittest.mock`, at least for unittests. Other than that, all you need to get started is this:

```python
def test_something():
    assert 1 == 2
```

You already suggested opening a ticket for missing tasks, perhaps we can just do that, use it as a meta-ticket to collect tasks and discuss things related to tests. For example, if you have ideas for tests you can sketch them out and I can help. What timezone are you in? I think it's going to be difficult to find time where we could talk things through in realtime, but it'll work.

squiddy:
@AndyMender What python version are you using locally? My guess is that Python 3.4, as used on travis, triggers the failure with the status icons easier.

AndyMender:
I had a look at the pytest docs and it's actually a lot more streamlined and straightforward than I initially expected. As you noticed, it really boils down to a simple `test_*` function definition with an `assert` evaluation in it. I actually do have ideas for some unittests so I'll just go ahead and write them, though your feedback would be more than welcomed, of course. :)

I use Python 3.6 on openSUSE Tumbleweed, though I can sacrifice myself and run Leap 42.2 instead to be closer to what Travis CI runs. Still, I feel it's a bit disturbing Travis CI uses an outdated environment such as Ubuntu 14.04 LTS. Everything else out there already has at least Python 3.5.x.

AndyMender:
@squiddy, you were right about Python 3.4.x being the culprit of test_inventory_full failing. I can reproduce the issue on a fresh install of openSUSE Tumbleweed 42.2, which happens to have Python 3.4.5. I will investigate this.

MasterofJOKers:
Just to chime in: Debian jessie (which is still the current "stable") doesn't have Python 3.5 either. So if you feel like it, install a Debian jessie chroot with debootstrap.

squiddy:
Thanks guys for your input. Very much appreciated.

@AndyMender While 3.4 is old, iirc it's the minimum python version we decided to support, didn't we?

We could merge master into this branch, but I'd rather not make this a long standing branch. It's really this one issue that I consider a blocker.

squiddy:
> EDIT: It seems like the order of tests in tests/game/test_status_icon.py matters. I repositioned assert_called_with_icon to the very end of the file and then only the unhappiness test failed.

I'll definitely have to check that out. It seems counter-intuitive, since `assert_called_with_icon` is just a helper function and not a test, therefore the order should be the same. If you run `pytest -v tests/game/test_status_icons.py` it'll show more information, such as the order of tests executed.

With the `pytest-xdist` plugin (which allows running tests in parallel) one could try to use the `boxed` feature (where each test is ran isolated in a separate process) like this: `pytest --boxed -v tests/game/test_status_icons.py`. This should answer the question of whether the tests somehow influence each other (which they should not).

AndyMender:
> Just to chime in: Debian jessie (which is still the current "stable") doesn't have Python 3.5 either. So if you feel like it, install a Debian jessie chroot with debootstrap.

When aiming for "Jessie" initially I noticed that there were some issues with library compatibility out-of-the-box (I think either swig or cmake) for building FIFE on Debian Stable. I selected Python 3.4 as the minimal requirement, because then it was the main Python3 version on most other systems, too. The problem with Python is that it moves way too fast. If we consider going for an official Debian/Ubuntu package, Python 3.5 will quickly become the target version.

Still, I would like to make sure that Unknown Horizons is compatible with at least Python 3.4 for the time being.

MasterofJOKers:
jessie has swig3.0 and cmake 3.6 in jessie-backports (3.0.2 in jessie itself, but fife needs 3.2.2).

LinuxDonald:
i´m happy when debian 9.0 is out (stretch)

squiddy:
Python 3.5 doesn't help with the issue at hand because I can also reproduce it when running tests in parallel. So whether or not 3.4 is the right version to support isn't relevant to this PR.

AndyMender:
You are right. Also, I just ran `pytest -v tests/game/test_status_icons.py` and the output is clean. All tests passed.

LinuxDonald:
rebased.

AndyMender:
I updated the gui test fixtures: df7a14e2477527e54db90580b3f024ef35e7b0c5; Here is the regular gui test run from master: https://pastebin.com/L1fBgqxV
Apart from the recent mishap with string templates, it doesn't look too bad and we can proceed with rewriting the tests for use with `pytest`.

AndyMender:
@squiddy, since my pull request and yours cover similar parts of the code base, could we merge them together somehow? :)

squiddy:
@andymender I'd rather not, the PR would become too big. Let me have a look this evening.

squiddy:
I try to get that one issue fixed so this can be merged instead.

squiddy:
> After attempting to track down the random failed assert calls, I noticed that AddStatusIcon.subscribe might be the culprit. This touches on the message bus system established by scripts in horizons/messaging, relying on some sort of a global state. Unfortunately, this is out of my league for now :(.

@AndyMender Great detective work! I've just installed python 3.4, and that seems very likely to be the problem. I'm testing a patch right now, looking good! Will push later today.

squiddy:
Hurray, the test run succeeded. However, there is one more thing I have to do. That has to wait until tomorrow evening though, but I'll have more time then. So I'm definitely going to merge this tomorrow. :)

AndyMender:
I ran pytest in your branch several times and each time it passed. Very good job, indeed! :)

squiddy:
Merged in fa2d0539b5e74ee99d8dadf0a3e1d463906d52a2

PR #2669, opened 2017-05-09
2667 imp to importlib

AndyMender:
Replace the deprecated `imp` module with a workaround that temporarily modifies `sys.path`. This is a simpler solution than using `importlib` (replacement for `imp`) or third-party modules.

### DIFF
--- a/run_uh.py
+++ b/run_uh.py
@@ -279,6 +279,8 @@ def find_fife(paths):
 	"""Returns True if the fife module was found in one of the supplied paths."""
 	default_sys_path = sys.path # to restore sys.path later
 	for path in paths:
+		if path.endswith("fife"):
+			path = os.path.dirname(path)
 		sys.path.insert(0, path)
 		try:
 			import fife
--- a/run_uh.py
+++ b/run_uh.py
@@ -279,7 +279,8 @@ def find_fife(paths):
 	"""Returns True if the fife module was found in one of the supplied paths."""
 	default_sys_path = sys.path # to restore sys.path later
 	for path in paths:
-		if path.endswith("fife"):
+		# extract parent directory to FIFE module
+		if path.endswith("fife") and os.path.isdir(path):
 			path = os.path.dirname(path)
 		sys.path.insert(0, path)
 		try:
--- a/run_uh.py
+++ b/run_uh.py
@@ -285,17 +285,22 @@ def find_fife(paths):
 		sys.path.insert(0, path)
 
 		try:
-			from fife import fife
-			break
-		except ImportError as e:
+			import fife
+			try:
+				from fife import fife
+				break
+			except ImportError as e:
 				if str(e) != 'cannot import name fife':
 					log().warning('Failed to use FIFE from %s', fife)
 					log().warning(str(e))
 					if str(e) == 'DLL load failed: %1 is not a valid Win32 application.':
 						# We found FIFE but the Python and FIFE architectures don't match (Windows).
 						exit_with_error('Unsupported Python version', '32 bit FIFE requires 32 bit (x86) Python 3.')
+				return False
+		except ImportError:
+			pass
 	else:
-		sys.path = default_sys_path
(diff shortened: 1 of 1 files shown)


run_uh.py
```diff
@@ -277,22 +277,36 @@ def flush(self):
...
+		log().warning("Module FIFE not found in supplied paths: {}.".format("\n".join(paths)))
+
+	# restore sys.path
+	sys.path = default_sys_path
+
+	# check if fife import was registered
+	if "fife" not in sys.modules:
+		return False
+
+	try:
```
squiddy: I'm wondering if this block can replace line 287? That way we're only attempting the import once.

That way, we'd only be replacing these two lines basically:

```python
module_info = imp.find_module('fife', paths)
fife = imp.load_module('fife', *module_info)
```

AndyMender: Not sure if I fully understand, but as far as I know, sys.modules contains the name of the module only after it has been imported. Therefore, we need to attempt to import it first :).

squiddy: Yes, I think we shouldn't be restoring sys.path at all if the import succeeded. What I meant was:


```python
	default_sys_path = sys.path # to restore sys.path later
  	for path in paths:
  		# extract parent directory to FIFE module
  		if path.endswith("fife") and os.path.isdir(path):
  			path = os.path.dirname(path)
  		sys.path.insert(0, path)
		try:
			from fife import fife
			break
		except ImportError as e:
			if str(e) != 'cannot import name fife':
				log().warning('Failed to use FIFE from %s', fife)
				log().warning(str(e))
				if str(e) == 'DLL load failed: %1 is not a valid Win32 application.':
					# We found FIFE but the Python and FIFE architectures don't match (Windows).
					exit_with_error('Unsupported Python version', '32 bit FIFE requires 32 bit (x86) Python 2.')
	else:
		sys.path = default_sys_path
		return False

	return True
```

AndyMender: Yes, something along these lines might work :).

AndyMender:
I think handling `from fife import fife` directly might give us confusing results - `except ImportError` will be triggered already when the original `fife` module doesn't exist and trying to log the underlying `fife` will cause an exception within the `except` handler :(.

squiddy:
> except ImportError will be triggered already when the original fife module doesn't exist [...]

I understand that this happens, but what do we gain by logging the `fife` variable? That's what is showing up if I just `print(fife)` after `import fife`:

`<module 'fife' from '/home/reiner/.pyenv/versions/uh/lib/python3.6/site-packages/fife/__init__.py'>`

We might as well just print the path we attempted to import from, in this case this would be `/home/reiner/.pyenv/versions/uh/lib/python3.6/site-packages/`.

AndyMender:
I personally know of cases when a direct `fife` import will succeed, but `from fife import fife` will fail due to issues with swig converting the _fife.so into fife.py. The 2-step approach helps catch those.

squiddy:
> The 2-step approach helps catch those.

And why wouldn't the 1-step approach do as well? You only mentioned it triggers an exception when logging `fife`. Is there something else?

AndyMender:
Because if the `fife` module cannot be loaded and you do `from fife import fife` and later use the underlying `fife` object, this will break the `except` block itself. Is this correct?

squiddy:
Yes, but **why** do you want to use the underlying `fife` object? Yes, the current code logs it, but I was asking if this is really necessary.

AndyMender:
I cannot judge whether it is necessary outside of the scenario I mentioned (`fife` is loaded, but `from fife import fife` fails due to problems with swig). We even handle errors in getting `fife` from `fife` so maybe such problems existed.

squiddy:
I think we're talking past each other, not sure how to resolve that. I might just test it myself. :)

squiddy:
@andymender No, no. I don't want to rely on feelings, I haven't tested anything. We keep it that way, I'd like to have a test here but that's going to be a bit easier once pytest is merged.

I think this can be merged if you tested it. But please rebase your changes into a single commit first.

PR #2675, opened 2017-05-13
Additional unittests with pytest (depend on pr #2668)

AndyMender:
This is an attempt to write extra tests for the functionalities with missing or incomplete coverage using the `pytest` library. Here is a quick report on the coverage level as produced by running `COVERAGE_FILE=.coverage.nongui pytest --verbose --cov`: https://pastebin.com/5jxZDfaC


squiddy:
When opening a pull request you can select the branch to merge into. This definitely shouldn't be master since we have my pr for that. So you could instead base this on the pytest branch. Also, does this contain any new commits?

AndyMender:
No new commits yet. Just wanted to set things up first and possibly discuss which sections of our code base require extra attention.

squiddy:
Why would we need a pull request for that? And why do you request a review from me. I don't understand this. You can set things up locally just fine.

squiddy:
Also I'm going to rebase the pytest branch onto master soon, so all that "merge master" noise goes away.

AndyMender:
What do you mean by "set things up locally"? I do not understand. Also, it is entirely unclear to me which specific functionalities in our code base require tests to be written. There is no comprehensive list anywhere. I wanted to discuss that as well. Not sure how am I supposed to do it otherwise if I never see you on the IRC channel...

LinuxDonald:
When squidds have done an rebase of his branch this branch need than to an rebase!

LinuxDonald:
rebased

squiddy:
@andymender If you want to discuss things an issue is better suited in my opinion. The thing is, my time is limited. If you request people's review they (at least I did) get a notification. Then I go check this out and there is nothing to do. No code, no questions. That was a bit annoying.

Regarding tests, have a look at the code coverage. While 100% coverage isn't the final goal, it can't hurt to have it at first step, especially since we have no type safety. Also, high level tests (game and gui) are abstract enough to not break often, as compared to specific unit tests.

Also, perhaps you already fixed bugs in the project. Check whether the features are covered by tests, otherwise they break again. Think about what we did with the cmdline options.

In my opinion there isn't really a wrong test right now. If it turns out to break too often (e.g due to refactoring) we can delete it. Still you gained experience with the test infrastructure and pytest.

squiddy:
TL;DR write game and gui tests to increase coverage, as they tend to break less often than unittests

squiddy:
Just to be clear, how I'm using github:

PR - here is some code I want you to merge / look at
Issue - let's talk about / discuss something

squiddy:
I'll have a look this evening.

tests/game/test_buildings.py
```diff
@@ -173,21 +174,58 @@ def test_build_tear(s, p):
...
 @game_test(timeout=60)
-def test_tree_production(s, p):
-	"""Check whether trees produce wood"""
+def test_tree_wood_production(s, p):
+	"""
+	Check whether trees produce wood
+	"""
 	settlement, island = settle(s)
+
+	# create building tree at coordinates (30,35)
```
squiddy: That comment doesn't add anything. I don't claim to be an export when writing comments, but typically comments should explain the why, and not the how.

This one says the same as the code below (when I strip some syntax). `Build tree at 30, 35`

tests/game/test_buildings.py
```diff
@@ -173,21 +174,58 @@ def test_build_tear(s, p):
...
 	tree = Build(BUILDINGS.TREE, 30, 35, island, settlement=settlement)(p)
 
+	# number of wood units to produce
```
squiddy: I'd drop the comment and the definition of `n` entirely, and just inline the number in the for loop.

`for i in range(20): # we want to produce 20 units of wood`

tests/game/test_buildings.py
```diff
@@ -173,21 +174,58 @@ def test_build_tear(s, p):
...
 	n = 20
 
+	# create inventory for building tree
```
squiddy: This doesn't actually create anything, it just gets a reference to the inventory. I'd drop the comment.

tests/game/test_buildings.py
```diff
@@ -173,21 +174,58 @@ def test_build_tear(s, p):
...
+
+
+@game_test(timeout=60)
+def test_tree_wildanimalfood_production(s, p):
+	"""
+	Check whether trees produce wild animal food
+	"""
+	settlement, island = settle(s)
+
+	# create building tree at coordinates (30,35)
```
squiddy: See comments for the other test.

tests/unittests/test_main.py
```diff
@@ -204,16 +204,34 @@ def test_load_game_by_name(self):
...
 
 
-@skip_todo
-def test_load_game_by_path(self):
-	pass
+@mock.patch('horizons.main.start_singleplayer')
+def test_load_game_by_path(mock_start_singleplayer):
+	"""
+	Test that a specific savegame file given by path can be loaded from the command line.
+	An in-game map is used instead of an actual savegame file - see NOTE above.
```
squiddy: Ok, my NOTE might be confusing. It wasn't meant to be a reason why we're going to do the tests differently, but rather explain why the tests haven't been implemented yet.

In this case, we don't have to commit a savegame/map specifically for the tests, we should avoid that if possible. Here, we want to test that loading a file by a path is possible, so a temporary file will work just fine. Sketch:

```python
with tempfile.NamedTemporaryFile() as f:
  start_game("--load-game", f.name)
  ...
  assert options.game_identifier == f.name
```
(Note: you probably have to add the `suffix` argument to NamedTemporaryFile)

Any file will do for now, since in the code under test it isn't even inspected. If that code ever changes, the test will break and we can adapt it. But for now, a random empty file is fine.

tests/unittests/test_main.py
```diff
@@ -204,16 +204,34 @@ def test_load_game_by_name(self):
...
-
-@skip_todo
-def test_edit_game_map_by_path(self):
-	pass
\ No newline at end of file
+@mock.patch('horizons.main.start_singleplayer')
+def test_edit_game_map_by_path(mock_start_singleplayer):
+	"""
+	Test that a specific savegame file given by path be loaded into the editor from the command line.
+	An in-game map is used instead of an actual savegame file - see NOTE above.
```
squiddy: See above.

tests/unittests/test_main.py
```diff
@@ -204,16 +204,34 @@ def test_load_game_by_name(self):
...
+	"""
+	start_game("--edit-game-map", "content/save/savegame_test.sqlite")
```
squiddy: Once NamedTemporaryFile is used, we don't need `content/save/savegame_test.sqlite` anymore. It's also a bit confusing to ship a savefile in the game's content directory.

AndyMender: I actually didn't know about `tempfile`, but just read the docs and the module seems to be extremely useful.

squiddy: Ah, what I did when I started with python is just look through most of the modules in the standard library. There are lots of nice things in there, e.g. itertools, collections, futures.

There was also a blog somewhere that regularly presented modules from the standard library with small examples.

AndyMender:
@squiddy, I applied your fixes, but skipped the final test, because it exposes more severe problems with our command-line interface for editing savegame files as maps. I will try to adjust it and run the tests again.

horizons/main.py
```diff
@@ -557,13 +557,10 @@ def edit_game_map(saved_game_name):
 	@return: bool, whether loading succeeded
 	"""
 	saved_games = SavegameManager.get_saves()
-	saved_game_path = _find_map(saved_game_name, saved_games)
-	if not saved_game_path:
+	map_name = _find_map(saved_game_name, saved_games)
+	if not map_name:
 		return False
 
-	accessor = SavegameAccessor(saved_game_path, False)
```
squiddy: Are you sure this is a fix? Isn't the idea of `--edit-game-map` to pass it a savegame, and use the map from that savegame in the editor? Does that still work?

I haven't tested it though, just curious why we wouldn't need to "parse" the SaveGameAccessor anymore.

AndyMender: The problem with the `accessor` is that if you try to load a savegame map that is not an actual savegame, the `accessor` will complain that the revision is 0 and raise `SavegameTooOld`. Even if this is circumvented/handled, the savegame metadata will still be missing for such a map and loading it will fail. 

On top of this, when I used an actual savegame file (the previous test_savegame.sqlite), it would be parsed by the `accessor` and saved in the database under a different name (either "development" or something else), possibly because as it is now, loading a map into the editor just loads the map's template.

I agree however, that it might be better to `todo_skip` this test or let it fail with the temporary file generated from `NamedTemporaryFile`, rather than modify map loading functions.

I ran GUI tests and the pytest tests after the modification and nothing was explicitly broken.

squiddy: > I ran GUI tests and the pytest tests after the modification and nothing was explicitly broken.

I doubt that the GUI tests even cover that functionality. Especially not the command line. I mean, we just started writing tests for all the flags.

> [...] if you try to load a savegame map that is not an actual savegame, the accessor will complain that the revision is 0 and raise SavegameTooOld

That sounds like reasonable behaviour. If the flag expects an actual savegame, and the test/user doesn't supply one, that feature is supposed to throw an error.

> I agree however, that it might be better to todo_skip this test or let it fail with the temporary file generated from NamedTemporaryFile, rather than modify map loading functions.

Yes, please. Because the way it sounds is that we're changing the implementation to make a test work. :+1: for Skip.

squiddy: > [...] possibly because as it is now, loading a map into the editor just loads the map's template.

Yes, I think that's the way this feature should work. And if we have a test that confirms this behaviour, we're done. Simple as that.

squiddy:
Did you have time to look into git rebase? Because over time, changing things and changing them back again adds unnecessary noise in the review, and later, when trying to find the source of bugs (more commits to look at).

AndyMender:
> Did you have time to look into git rebase? 
Not yet properly. Will do that soon :).

squiddy:
> Not yet properly. Will do that soon :).

In that case, I'd like to take over the merge for this branch once it is ready (if you don't feel familiar enough with rebase at that point). I believe currently you worked on maybe 4 tests (?) over the course of 16 commits.

AndyMender:
I fully agree! Will leave the final clean-up in your hands then :). Thank you again for the great feedback!

squiddy:
Damn, stupid GitHub auto-closed the PR because I deleted the pytest branch. -.- Sorry.

AndyMender:
That's alright :). I saw you opened a branch called pytest-gui. It is for the GUI tests refactoring, correct?
For this PR I will then continue working on new tests as originally intended :).

squiddy:
> I saw you opened a branch called pytest-gui. It is for the GUI tests refactoring, correct?

Yes, although I'll probably work on game tests first. Need to figure out how to fit the whole decorator logic better into pytest. Making good progress though.

> For this PR I will then continue working on new tests as originally intended :).

Yes, sounds good. Perhaps I'm going to clean it up and merge this evening so it's doesn't become too big. :)

squiddy:
Squashed all changes into these three for each test:

https://github.com/unknown-horizons/unknown-horizons/commit/de4c9f3f0799d8b795792b0430a587b292bbb846
https://github.com/unknown-horizons/unknown-horizons/commit/1d57db3ee36dfdf2431e8f29b5bee16aaf8484df
https://github.com/unknown-horizons/unknown-horizons/commit/ea21ac342686f1bec0ea86c7fd106cf304e59094

PR #2680, opened 2017-05-15

MarkusHackspacher:
change to the .format() function in horizons/ai/aiplayer/ files

### DIFF
--- a/horizons/ai/aiplayer/resourcemanager.py
+++ b/horizons/ai/aiplayer/resourcemanager.py
@@ -488,7 +488,7 @@ class SingleResourceManager(WorldObject):
 	def __str__(self):
 		if not hasattr(self, "resource_id"):
 			return 'UninitializedSingleResourceManager'
-		result = 'Resource {:d} production {:.5f}/{:.5f] ({:.5f} low priority)'.format(
+		result = 'Resource {:d} production {:.5f}/{:.5f} ({:.5f} low priority)'.format(
 			self.resource_id, self.available, self.total, self.low_priority)
 		for quota_holder, (quota, priority) in self.quotas.items():
 			result += '\n  {}quota assignment {:.5f} to {}'.format(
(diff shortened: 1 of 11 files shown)


horizons/ai/aiplayer/resourcemanager.py
```diff
@@ -487,9 +488,11 @@ def get_total_export(self):
 	def __str__(self):
 		if not hasattr(self, "resource_id"):
 			return 'UninitializedSingleResourceManager'
-		result = 'Resource %d production %.5f/%.5f (%.5f low priority)' % (self.resource_id, self.available, self.total, self.low_priority)
+		result = 'Resource {:d} production {:.5f}/{:.5f] ({:.5f} low priority)'.format(
```
AndyMender: There is an error here. Look closely at the curly brace pairing:
```py
'Resource {:d} production {:.5f}/{:.5f] ({:.5f} low priority)'.format()
```

horizons/ai/aiplayer/villagebuilder.py
```diff
@@ -854,4 +854,6 @@ def display(self):
 			renderer.addColored(tile._instance, *color)
 
 	def __str__(self):
-		return '%s VillageBuilder(%s)' % (self.settlement_manager, self.worldid if hasattr(self, 'worldid') else 'none')
+		return '{} VillageBuilder{})'.format(
```
AndyMender: You missed 1 brace here, should be:
```py
return '{} VillageBuilder({})'.format(
```

horizons/ai/aiplayer/resourcemanager.py
MarkusHackspacher: Is solve.

AndyMender:
Approved. For the future, I think it would be a good idea to give meaningful names to commits so that later we know what each of them did :).

PR #2685, opened 2017-05-17

MarkusHackspacher:

### DIFF
--- a/horizons/ext/speaklater.py
+++ b/horizons/ext/speaklater.py
@@ -200,7 +200,7 @@ class _LazyString:
         try:
             return 'l' + repr(self.value)
         except Exception:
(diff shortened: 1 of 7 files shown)


horizons/ext/speaklater.py
```diff
@@ -200,7 +200,7 @@ def __repr__(self):
         try:
             return 'l' + repr(self.value)
         except Exception:
-            return '<%s broken>' % self.__class__.__name__
```
squiddy: This is an external library, and therefore shouldn't be modified as it doesn't fall under our codestyle. This change should be reverted.

MarkusHackspacher: I have it revert.

Issue #2684, opened 2017-05-17
Use new version info hosted at uh.org for version hint


squiddy:
We're currently using the information at `http://updates.unknown-horizons.org/current_version.php`. To be more independent and to make updating that easier, it will be hosted at our homepage (see https://github.com/unknown-horizons/unknown-horizons.github.io/issues/8).

squiddy:
Because the file is now hosted on our static homepage, I'll proceed to change the update info feature accordingly.

PR #2688, opened 2017-05-18
Point update check to new file location and refactor

squiddy:
Closes: #2684

### DIFF
--- a/horizons/constants.py
+++ b/horizons/constants.py
@@ -28,6 +28,7 @@ from pathlib import Path
 from typing import List, Optional
 
 from horizons.ext.enum import Enum
+from horizons.util.platform import get_user_game_directory
 
 
 """This file keeps track of the constants that are used in Unknown Horizons.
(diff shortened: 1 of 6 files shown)


horizons/constants.py
```diff
@@ -635,7 +635,7 @@ class NETWORK:
 	# change port to 2022 for development server updated after UH commits
 	SERVER_PORT = 2002
 	CLIENT_ADDRESS = None # type: Optional[str]
-	UPDATE_FILE_URL = "http://updates.unknown-horizons.org/current_version.php"
+	UPDATE_FILE_URL = "http://unknown-horizons.org/current-version.json"
```
squiddy: It doesn't. The certificate for that domain is for `*.github.com`. According to [this](https://github.com/blog/2186-https-for-github-pages) GitHub Pages supports HTTP, but only for user pages. If I try and access `https://unknown-horizons.github.io/current-version.json` I get redirected to `http://unknown-horizons.org/current-version.json`.

squiddy: At least not directly with GitHub. I wonder if we could setup something before it, Cloudflare maybe? Then we could use `https://unknown-horizons.github.io`

horizons/util/checkupdates.py
```diff
@@ -19,99 +19,102 @@
...
-		if version != VERSION.RELEASE_VERSION:
-			# there is a new version
-			info.version = version
-			info.link = link
-			info.status = UpdateInfo.READY
-		else:
-			info.status = UpdateInfo.INVALID
-	else:
-		info.status = UpdateInfo.INVALID
+	# only updates for operating systems missing a package management
```
squiddy: Interesting idea for sure.

One potential issue for people running from git, that I see, is that it might be a developer, and if we were to check for new commits (maybe tags?) you'd get the popup very often. But perhaps you had something else in mind?

MasterofJOKers: I did not think about checking for git commits, but rather keep the check for the current version against the latest release. Even when running from git, I'd like to know if my version is outdated.

Imho, we can assume, that someone doesn't use package management, if UH is somewhere below the home-directory of the user.

MasterofJOKers: A developer probably doesn't get the popup, as he/she is running a newer/the latest version.

squiddy: Ah, gotcha. I can implement that while I'm at it.

PR #2693, opened 2017-05-23
Add more pytest-based building tests

AndyMender:
This is an attempt to write additional game tests for building and collector mechanics.
Most buildings are covered, except for the boat builder and the barracks test is set to `xfail` for now.


squiddy:
I'm going to rebase the branch onto master and give you a review then, later this evening.

AndyMender:
Thank you kindly :).

tests/game/test_buildings.py
```diff
@@ -29,6 +29,8 @@
 from horizons.world.production.producer import Producer
 from tests.game import game_test, settle
 
+import pytest
```
squiddy: This import is in the wrong order.

Try to make it a habit of running pycodestyle and isort against your changes. It should be between itertools (stdlib) and horizons (our code) imports.

tests/game/test_buildings.py
```diff
@@ -218,3 +220,393 @@ def test_tree_wildanimalfood_production(s, p):
...
+	pavilion = Build(BUILDINGS.PAVILION, 30, 30, island, settlement=settlement)(p)
+
+	assert pavilion
+	assert pavilion.get_component(StorageComponent).inventory[RES.FAITH] == 0
+	s.run(seconds=30)
+	assert pavilion.get_component(StorageComponent).inventory[RES.FAITH]
+
+
+@game_test()
+def test_farm_production(s, p):
```
squiddy: We already have tests for the farm in `tests/game/test_farm.py`. It has a nice utility function to build a farm with fields, those tests can probably be combined.

This test has the advantage that it tests more production chains, so we can combine the tests. If this test were to use the `_build_farm` function it would be better (and the TODO could be removed).

tests/game/test_buildings.py
```diff
@@ -218,3 +220,393 @@ def test_tree_wildanimalfood_production(s, p):
...
+	
+	assert saltpond
+	
+	assert saltpond.get_component(StorageComponent).inventory[RES.SALT] == 0
+	s.run(seconds=60)
+	assert saltpond.get_component(StorageComponent).inventory[RES.SALT] >= 2 # ponds produce salt in units of 2
+	
+
+@game_test()
+def test_distillery_production_chain(s, p):
```
squiddy: `tests/game/test_farm.py`  already has a distillery test, both are similar, so I guess just remove the one from that file. See comment above.

tests/game/test_buildings.py
```diff
@@ -20,6 +20,7 @@
 # ###################################################
 
 from itertools import product
+import pytest
```
squiddy: That's interesting, I would have thought pytest enforces a new line between groups. Weird.

AndyMender: This time I checked isort locally before committing. It was A okay.

squiddy: Yeah, Travis doesn't complain either. Nothing to worry though, I was just curious.

tests/game/test_buildings.py
AndyMender: Unfortunately, `_build_farm` for now would create 4 fields of the same type. I will see how to modify the function in such a way that more fields of different types can be placed next to one or more farms.

squiddy: Well, you don't *have* to use the function. Keep it simple. If `_build_farm` isn't a good fit, so be it.

I just want to make sure we don't have farm tests in two different files.

AndyMender:
In the end I expanded the `_build_farm` function (up to 8 fields) and used it to write 2 farm-based tests. `pycodestyle` was A okay and tests passed locally. I believe we can now remove the redundant test_farm.py test file.

tests/game/test_buildings.py
```diff
@@ -31,6 +31,30 @@
 from tests.game import game_test, settle
 
 
+def _build_farm(x, y, island, settlement, owner, *field_type):
+	"""
+	Build a farm at (x, y) and 9 fields of field_type.
+
+	F F F
+	F X	F	  (X - farm, F - field)
```
ChrisOelmueller: This tab should probably be a space ;-)

AndyMender: Yes, indeed. You are corrected. Fixed!

LinuxDonald:
@AndyMender status on this pr?

AndyMender:
@squiddy, is this good to go? All tests run fine locally.

squiddy:
@andymender Travis is reporting problems found by isort

AndyMender:
Now it should be fine. `isort` doesn't seem to like long `from module import object, ...` it seems.

squiddy:
I haven't read through it line by line, but it looks good.

PR #2689, opened 2017-05-18
change to the .format() function

MarkusHackspacher:
change to the .format() function.

PR #2694, opened 2017-05-23
Linechange

MarkusHackspacher:
See last RP #2689


MarkusHackspacher:
There was unsolve problems, I make a new one.

Issue #1379, opened 2012-03-27
Fix action set offsets


totycro:
The player ship and others are misplaced by a few pixels, one can see it when a ship is near the coast and the view is rotated. This can probably be fixed by correcting the offset in `units/__init__.py:65`.

Also check buildings, it looks a bit odd when the surrounding is highlighted (the buildings aren't exactly above the non-highlighted tiles below it.

![bug-1370](https://cloud.githubusercontent.com/assets/50333/19413544/86501c2a-9330-11e6-916f-54f7c8bd3237.png)
- [ ] Fix offsets for every rotation in ships
- [x] Check which buildings seem to hover above ground when activating tile highlights
- [x] Fix building offsets

LinuxDonald:
@stubb what about this issue?

LinuxDonald:
@stubb in what file must the offset corrected?

squiddy:
Updated the issue with a screenshot. @LinuxDonald If I had to guess, it's this line https://github.com/unknown-horizons/unknown-horizons/blob/8049731/horizons/world/units/__init__.py#L92

The `:shift:center+0,bottom+8` part. We could adjust that for every angle until it works in all rotations.

LinuxDonald:
Must I change only the numbers in that line?

MasterofJOKers:
Well yes and no. We need to change the numbers in that line based on the rotation. My guess would be, that it's always 8, but sometimes center+8, then center-8 and bottom-8 ... but that's just a guess. If you only change the numbers in that line, it should only look correct in one rotation, I guess.

squiddy:
Buildings have an issue as well:

![uh-bug-1379](https://cloud.githubusercontent.com/assets/50333/19416094/c099d022-9387-11e6-875b-c6cd0e9ed675.png)

On the right it almost looks like the building is hovering above ground. When activating the grid, it seems like grid and building are aligned correctly, but don't seem to match the tiles, they should be moved down a bit.

squiddy:
> yeah, part of that ticket is investigating how the image size (and proportions) play into this fife stuff i guess

> like, if i use a 256x256 image, or a 192x192 image, or a 128x128 image all for the same "2x2" building, what does change

-- eoc on irc

LinuxDonald:
Charcoal Burning is broken too. 
[...]

PR #2705, opened 2017-06-06
Fix smeltery and charcoal_burning position offsets

AndyMender:
I added extra *hacks* for the smeltery and charcoal burner buildings. Now they are properly positioned. Although this resolves issue #1379, I feel our offset system is inherently broken and any semi-manual x , y adjustments barely compensate for that.


AndyMender:
> I haven't checked these changes in-game, but I trust that you did. :)
Yes, I actually had to run UH multiple times to get the offsets as good as possible.
I think the dict you sketched might be a good idea in the future, thanks! :)

Issue #1678, opened 2012-04-12
Show translation completeness stats


otinn:
It could be useful to show how much of a translation is complete in the game. This could be separate for specific scenarios and the game in general. It could be shown in the language selection and scenario language selection menus.
##

PR #2709, opened 2017-06-08
#1678 Show translation completion status in settings menu

squiddy:
When building mo files, translation stats for each processed language are stored in a json file that will be parsed when the language is changed in the menu.

2017-06-08-221142_1314x834_scrot

Translation stats visible below the list of languages.


squiddy:
It's missing the update if translation files, I need to complete work on #2710 for that.

AndyMender:
Looks good to me. I like the idea of providing users with information on translation completion.

squiddy:
Merged.

PR #2712, opened 2017-06-09
Use `is` or `is not` to compare with `None`

MarkusHackspacher:
Use `is` or `is not` to compare with `None`

https://www.quantifiedcode.com/app/issue_class/3IY8CZ0v


squiddy:
Ah yes, good catch.

PR #2713, opened 2017-06-10
optimized test_namedcomponent

MarkusHackspacher:
- new def setup for reduce calls
- add comments
- call as staticmethod

### DIFF
--- a/tests/unittests/world/component/test_namedcomponent.py
+++ b/tests/unittests/world/component/test_namedcomponent.py
@@ -25,23 +25,34 @@ from horizons.component.namedcomponent import NamedComponent
 
 
 class MockNameComponent(NamedComponent):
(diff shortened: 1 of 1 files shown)


MarkusHackspacher:
I have no experience with pytest. Maybe you could write a short description (a README in test directory) how to make this change.

squiddy:
> I have no experience with pytest. Maybe you could write a short description (a README in test directory) how to make this change.

What I've done now (because I don't really want to talk about pytest's features here), is to rewrite these tests in a style more suitable for pytest. I've added some comments for you that talk about some of those features: 1bbd1f8c7fb6c02934fa3094afd25143d0e176e1

MarkusHackspacher:
looks fine with pytest.

PR #2722, opened 2017-06-19
Disable most military features

AndyMender:
Since we haven't yet implemented most of the combat features in Unknown Horizons, the possibility of erecting military buildings may be confusing to the players. I disabled them in the respective .yaml file, however left the cannon foundry still enabled as that's what I would like to focus on next.


squiddy:
If you want to focus on cannon foundry next, you can always reenable it in your branch. If you disable useless military buildings, be consistent about it.

Also, I expect a test to fail that makes sure that both build menu configurations are in sync. It seems you only updated one file.

AndyMender:
I have not looked at the second build_menu file, apologies. As you observed, the related test did fail. Now it's fixed, though.

LinuxDonald:
Cannon foundry itself works without problems. It build cannons but the can't used yet only equip on ship. But ship fight not works yet.

squiddy:
Please rebase these changes into a single commit before the merge.

squiddy:
> Cannon foundry itself works without problems. It build cannons but the can't used yet only equip on ship. But ship fight not works yet.

But that's the point: they can not really be used in any useful way.

AndyMender:
> Please rebase these changes into a single commit before the merge.

Did you mean squash? And yes, right now the naval combat system is pretty much non-existent also. I will open a new issue for that soonish.

squiddy:
> Did you mean squash?

Eh yes, squash. :)

PR #2732, opened 2017-07-16
Simplify explicit super() calls.

AndyMender:
This is an attempt to reduce the explicit calls `super(ChildClass, self).superclass_method()` to `super().superclass_method()`, as per changes in Python3 the `superclass_method` is bound to the instance of `ChildClass` by default. This does not touch cases when the `superclass_method` is bound to another object like in `super(ChildClass, cls).superclass_method()`.

In the midst of my analysis I noticed cases when a base class calls `super()` on itself as in:
```python 
class A:
     def __init__(self, x, y):
         super().__init__()
         self.x = x
         self.y = y
```
The Python interpreter claims it's wrong and at first glance may lead to infinite loops regardless. These cases were labeled with a TODO comment.


AndyMender:
Apparently, in cases when the `classmethod` decorator is used, my changes broke quite some tests. I will fix them soon.

AndyMender:
I fixed the broken `super()` calls, but the surrounding code that caused the issues is really ugly and should be fixed at one point in time.

LinuxDonald:
@AndyMender any news about that here?

AndyMender:
@LinuxDonald , all of the tests on Travis (including the GUI tests) run fine since the last commit. When I did manual game testing after that everything was fine also. However, I do need squiddy's feedback, as some use cases of `super()` are very non-obvious, especially in multiple inheritance scenarios and considering the fact that in some of the files even `__new__` was overloaded. Not to mention that our tests don't cover everything yet.

PR #2751, opened 2017-10-24
constants.py: set encoding

s172262:
`setup.py` caused the following errors.
```
Traceback (most recent call last):
  File "setup.py", line 34, in <module>
    from horizons.constants import VERSION
  File "/private/tmp/unknown-horizons-20171024-84852-cz9n25/horizons/constants.py", line 691
SyntaxError: Non-ASCII character '\xd8' in file /private/tmp/unknown-horizons-20171024-84852-cz9n25/horizons/constants.py on line 691, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

In http://python.org/dev/peps/pep-0263/
>Python will default to ASCII as standard encoding if no other encoding hints are given.

### DIFF
--- a/horizons/constants.py
+++ b/horizons/constants.py
@@ -1,3 +1,4 @@
+# This Python file uses the following encoding: utf-8
 # ###################################################
 # Copyright (C) 2008-2017 The Unknown Horizons Team
 # team@unknown-horizons.org


squiddy:
We switched to Python 3 a while back, which defaults to utf-8, so this shouldn't really be an issue, right? I don't have the time to test it though.

s172262:
I'm sorry. I used `Python2`.

LinuxDonald:
I hope that i can find out how to build fife on macos with python3. Because our next version will only support python3 on UH side.

PR #2755, opened 2017-11-17
Fix combat system

LinuxDonald:


horizons/world/units/fightingship.py
```diff
@@ -74,7 +77,7 @@ def act_attack(self, dest):
 		side = (x2 - x1) * (y3 - y1) - (y2 - y1) * (x3 - x1)
 		# calculate x4 y4 the new facing location coords
 		# they are calculated by rotating 90' the target location
-		if (side or 0) > 0:
```
squiddy: Are you *sure* this change is correct? IIRC we added that when we switched to Python3, because `side` was sometimes `None` - I think during tests.

LinuxDonald: Reverted it for now

PR #2757, opened 2017-11-24
Avoid double-firing of `keyPressed` events

FedericoStra:
The `IngameKeyListener` seems to double fire `keyPressed` events.
The net effect is that it is impossible to use hotkeys such as 'B'
because the dropdown construction menu just flashes for an instant
and then closes back down immediately.

I suspect this is a deeper issue inside `fifengine`, but I am not
familiar with that so here I offer a quick fix to avoid the problem.

Simply put, the `IngameKeyListener` keeps track of the last event
and, if the current one is identical (same key, same press/release
status), it directly discards it without further processing.

Furthermore, I also cache an instance of `KeyConfig` inside
`IngameKeyListener` instead of instantiating one every time
`keyPressed` or `keyReleased` are called. This just popped to my eyes
as an obvious optimization to implement, but feel free to discard if
it gives raise to some problems.

### DIFF
--- a/horizons/gui/keylisteners/ingamekeylistener.py
+++ b/horizons/gui/keylisteners/ingamekeylistener.py
@@ -35,6 +35,7 @@ class IngameKeyListener(fife.IKeyListener, LivingObject):
 		from horizons.session import Session
 		assert isinstance(session, Session)
 		self.session = session
+		self.keyconfig = KeyConfig()
 		horizons.globals.fife.eventmanager.addKeyListenerFront(self)
 		# Used to sum up the keyboard autoscrolling
 		self.key_scroll = [0, 0]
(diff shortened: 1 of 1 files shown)


FedericoStra:
Here is the associated issue: #2758

LinuxDonald:
@helios2000 could it be an fife/fifechan problem?

@FedericoStra Thank you for this pr :) First we will took on the issue whats going on.

LinuxDonald:
@AndyMender could you maybe review this please? :)

AndyMender:
Looks alright to me, but I too lack discrete knowledge on how Unknown Horizons handles key presses. @FedericoStra if you don't mind, we'll hold back with merging for now as our test server is in a shamble and many of the tests fail even locally. I would like to address that first.

PR #2764, opened 2017-12-28
Ifixthings

ttoocs:
The minimal number of changes to get things to work (as in, not crash. Not sure why there's parameters getting read/set that don't exist... but I just wanted to play the game)
TL:DR; I could play a game in singleplayer after this.

### DIFF
--- a/horizons/world/traderoute.py
+++ b/horizons/world/traderoute.py
@@ -127,7 +127,7 @@ class TradeRoute(ChangeListener):
 			for res in copy.copy(self.current_transfer):
 				# make sure we don't keep trying to (un)load something when the decision about that resource has changed
 				if self.current_transfer[res] == 0 or res not in self.get_location()['resource_list'] or \
-				   cmp(self.current_transfer[res], 0) != cmp(self.get_location()['resource_list'][res], 0):
+				   ((self.current_transfer[res] < 0) != (self.get_location()['resource_list'][res] < 0)): #If unload/load has changed, (based on signs)
 					del self.current_transfer[res]
 
 		settlement = warehouse.settlement
--- a/horizons/world/traderoute.py
+++ b/horizons/world/traderoute.py
@@ -126,7 +126,7 @@ class TradeRoute(ChangeListener):
 		if self.current_transfer is not None:
 			for res in copy.copy(self.current_transfer):
 				# make sure we don't keep trying to (un)load something when the decision about that resource has changed
-				if self.current_transfer[res] == 0 or res not in self.get_location()['resource_list'] or \
+				if self.current_transfer[res] == 0 or res not in self.get_location()['resource_list'] or self.get_location()['resource_list'][res] == 0 or \
 				   ((self.current_transfer[res] < 0) != (self.get_location()['resource_list'][res] < 0)): #If unload/load has changed, (based on signs)
 					del self.current_transfer[res]
 


LinuxDonald:
The problem is that tier5 is not yet correctly implemented and that's why I think that the crashes comes from. But we will check some changes from your pr.

ttoocs:
The fixes provided fix crashes when: 
The AI first makes a settlement - Runs into a chain.assign_identifier is undefined for NoneType, now checks if it's NoneType.
When the AI upgrades to tier2 - These properties are never found.. so I just commented them out.
And when you get a trade ship to arrive at your dock to pull resources. - Crashes when comparing something, saying the function is undefined. (A remnant from python2?)

I'm not sure if others have encountered these errors, but I encounter them on while playing on arch linux and fixed them so that I could play.  I hope it's helpful.

LinuxDonald:
@AndyMender @squiddy could maybe one of you review this please?

AndyMender:
I'll have a look at it today-tomorrow, but I have never experienced this issue myself so far.

LinuxDonald:
@AndyMender me too but maybe its an random error? We will see after your review what you about the changes think.

ttoocs:
Oh, didn't think it would include that with the pull request... /sigh.

horizons/ai/aiplayer/productionchain.py
```diff
@@ -53,8 +53,9 @@ def __init__(self, settlement_manager, resource_id, resource_producer):
 		self.settlement_manager = settlement_manager
 		self.resource_id = resource_id
 		self.chain = self._get_chain(resource_id, resource_producer, 1.0)
-		self.chain.assign_identifier('/{:d},{:d}'.format(
-			self.settlement_manager.worldid, self.resource_id))
+		if not self.chain is None:
```
AndyMender: Why not simply `if self.chain:`?

horizons/ai/aiplayer/settlementmanager.py
```diff
@@ -228,13 +228,15 @@ def _set_taxes_and_permissions_prefix(self, prefix):
...
 		sailor_upgrades = getattr(self.personality, '{}_sailor_upgrades'.format(prefix))
 		pioneer_upgrades = getattr(self.personality, '{}_pioneer_upgrades'.format(prefix))
 		settler_upgrades = getattr(self.personality, '{}_settler_upgrades'.format(prefix))
-		citizen_upgrades = getattr(self.personality, '{}_citizen_upgrades'.format(prefix))
-		self._set_taxes_and_permissions(sailor_taxes, pioneer_taxes, settler_taxes, citizen_taxes, merchants_taxes,
-			sailor_upgrades, pioneer_upgrades, settler_upgrades, citizen_upgrades)
+		#citizen_upgrades = getattr(self.personality, '{}_citizen_upgrades'.format(prefix))
+		merchants_taxes = None
+		citizen_upgrades = None
+		#self._set_taxes_and_permissions(sailor_taxes, pioneer_taxes, settler_taxes, citizen_taxes, merchants_taxes,
```
AndyMender: I don't think commenting out the entire call is a good idea. Notice that also ignores *_upgrades and *_taxes for valid tiers (sailors, pioneers, etc.).

`getattr()` also takes a default value as the last argument so that could be used to indicate that `merchants_taxes` and `citizen_upgrades` is missing / not implemented.

horizons/world/traderoute.py
```diff
@@ -31,6 +31,8 @@
 from horizons.util.worldobject import WorldObject
 from horizons.world.units.unitexeptions import MoveNotPossible
 
+def cmp(a, b):
```
AndyMender: Was this function missing previously? Could you attach the related traceback to the issue / PR?

horizons/ai/aiplayer/settlementmanager.py
ttoocs: You're definitely right.  I'm confused on how that occurred, as that is the entire point of setting the two values to None was to make it still possible to do that call. (Will test it again, in the rare cause calling it with None was breaking it.)

horizons/world/traderoute.py
ttoocs: I'm sorry, this was just found to be undefined in my own play of the game. (when I tried to setup a trade route, the game crashed, complaining that this function was not found)
- I didn't setup any issue, I just fixed it.

horizons/ai/aiplayer/productionchain.py
ttoocs: Done, seems to work. Thank you.

horizons/ai/aiplayer/settlementmanager.py
ttoocs: It was intentonal that I commented it out. There was an attempt with them set to None, (which is why they're set that way) but that crashes (as follows). Note: this isn't a complete fix, just one that gets it not-crashing.
`Traceback (most recent call last):
  File "./run_uh.py", line 385, in <module>
    main()
  File "./run_uh.py", line 127, in main
    ret = horizons.main.start(options)
  File "/home/fu-fu/source/unknown-horizons/horizons/main.py", line 255, in start
    horizons.globals.fife.run()
  File "/home/fu-fu/source/unknown-horizons/horizons/engine/engine.py", line 280, in run
    self.loop()
  File "/home/fu-fu/source/unknown-horizons/horizons/engine/engine.py", line 297, in loop
    f()
  File "/home/fu-fu/source/unknown-horizons/horizons/timer.py", line 112, in check_tick
    f(self.tick_next_id)
  File "/home/fu-fu/source/unknown-horizons/horizons/scheduler.py", line 95, in tick
    callback.callback()
  File "/home/fu-fu/source/unknown-horizons/horizons/util/python/callback.py", line 47, in __call__
    return self.callback(*self.args, **self.kwargs)
  File "/home/fu-fu/source/unknown-horizons/horizons/ai/aiplayer/__init__.py", line 317, in tick
    self.handle_settlements()
  File "/home/fu-fu/source/unknown-horizons/horizons/ai/aiplayer/__init__.py", line 364, in handle_settlements
    settlement_manager.refresh_taxes_and_upgrade_permissions()
  File "/home/fu-fu/source/unknown-horizons/horizons/ai/aiplayer/settlementmanager.py", line 376, in refresh_taxes_and_upgrade_permissions
    self._set_taxes_and_permissions_prefix('early')
  File "/home/fu-fu/source/unknown-horizons/horizons/ai/aiplayer/settlementmanager.py", line 239, in _set_taxes_and_permissions_prefix
    sailor_upgrades, pioneer_upgrades, settler_upgrades, citizen_upgrades)
TypeError: _set_taxes_and_permissions() takes 8 positional arguments but 10 were given
AL lib: (EE) alc_cleanup: 1 device not closed`

ttoocs: A better fix (I assume) is now provided, altering the method call to match the function.

horizons/world/traderoute.py
AndyMender: `cmp` is a Python 2 function, that's probably why it was undefined if you were running the game with Python 3. Perhaps a remnant of refactoring from Python 2? If it's used only once in the entire code base, I think it would be better to replace the functionality internally instead :).

ttoocs: It's used twice on line 134, not entirely sure the best method to clean this up.

nihathrael:
We fixed the root cause for the crash in the production chain in d66363bbf44b0a1976b6855736f418d3228cb2e3
The root cause for the tax management in the settlement manager was fixed in 8836e47

The cmp() still needs fixing though. I think it could be moved to `horizons/util/python/__init__.py ` or we try to get rid of it completely. Unfortunately, the code is not trivial to understand and hard to test. 

Can you adjust the PR accordingly? Thanks!

LinuxDonald:
@AndyMender what is the status here? About review

AndyMender:
@LinuxDonald Sorry about this. I somehow missed the updates to the PR.

jmdejong:
I don't think it's too convoluted.
There isn't really a way to simplify this if statement.
It would be possible to make this line a little shorter by adding more local variables or a function that does this, but that would just add more code elsewhere.

If you meant that the class is convoluted: it could be, but that's not really the responsability of this PR.
This is simply about fixing crashes

squiddy:
With a bit of formatting, the if would be a bit more readable:

```python
if (self.current_transfer[res] == 0 or 
        res not in self.get_location()['resource_list'] or 
        self.get_location()['resource_list'][res] == 0 or
        ((self.current_transfer[res] < 0) != (self.get_location()['resource_list'][res] < 0))): #If unload/load has changed, (based on signs)
    del self.current_transfer[res]
```

Also, these two lines can be further simplified:

```python
res not in self.get_location()['resource_list'] or 
self.get_location()['resource_list'][res] == 0 or
```

to

```python
self.get_location()['resource_list'].get(res, 0) == 0
```

jmdejong:
@squiddy good points. I implemented these changes in my own branch. I also used helper variables to make it look a bit cleaner.
See:
https://github.com/jmdejong/unknown-horizons/blob/d6095ffc24325e7dbbcdb7ce6d0e71ca78310672/horizons/world/traderoute.py#L127-L132
I PR'ed to ttoocs fork: https://github.com/ttoocs/unknown-horizons/pull/1

AndyMender:
```python
for res, current in self.current_transfer.copy().items():
	# make sure we don't keep trying to (un)load something when the decision about that resource has changed
	trade_stock = self.get_location()['resource_list'].get(res, 0)
	if current == 0 or trade_stock == 0 or (current < 0) != (trade_stock < 0):
		#If unload/load has changed, (based on signs)
		del self.current_transfer[res]
```
This looks quite good already, though 2 things:
- you probably don't need `copy()` when calling `items()`, since that already returns a view of the `dict`, which you can modify in the `if` block
- I would avoid the `del` statement - you can use the `dict.pop()` method; `del` can mess up garbage collection

jmdejong:
`copy()` is necessary to prevent errors. Items only gives a view, but the view depends on the original dictionary. as a test, try this:

    a = {1:1, 2:2, 3:3}
    for k,v in a.items():
      a.pop(k)

I never found anything about the `del` messing up garbage collection (at least when used on collections). As far as I know the only difference between del and pop is that pop returns the deleted value. But if you prefer pop I can change it.

Should I PR my version to master or should I wait for ttoocs to merge my PR into this PR?

AndyMender:
@jmdejong You can go ahead and submit a PR from your repository. I will review it and merge the changes.

PR #2765, opened 2018-01-04
Lots of test skipping added to make the tests pass.

ttoocs:
Not sure why all the tests are in a state of failing, as some should work at a glance (test_trivial in test_internal.py)
This should be fixed at somepoint.


ttoocs:
(Note: I don't yet want the pulled, I just need travis-CI to run on it.)

PR #2766, opened 2018-01-04
Add in fixme tests

ttoocs:
A better approach than simply skipping all the known-broken tests.
This adds a pytest mark 'fixme', and flags all currently broken tests with it.

### DIFF
--- a/horizons/main.py
+++ b/horizons/main.py
@@ -64,7 +64,7 @@ if TYPE_CHECKING:
 """
 Following are a list of global variables. Their scope is this module.
 Since this is not a class, in each function where these variables are
-referenced, they need to be declared with 'global' keyword. 
+referenced, they need to be declared with 'global' keyword.
 
 See: http://python-textbok.readthedocs.io/en/1.0/Variables_and_Scope.html
 """
@@ -596,7 +596,7 @@ def set_debug_log(enabled, startup=False):
 	@param startup: True if on startup to apply settings. Won't show popup
 	"""
 	global gui, command_line_arguments
- 
+
 	options = command_line_arguments
 
 	if enabled: # enable logging
--- a/horizons/view.py
+++ b/horizons/view.py
@@ -73,7 +73,7 @@ class View(ChangeListener):
 
 		rect = fife.Rect(0, 0, horizons.globals.fife.engine_settings.getScreenWidth(),
 		                       horizons.globals.fife.engine_settings.getScreenHeight())
-		self.cam = self.map.addCamera("main", rect)
+		self.cam = self.map.addCamera("main", self.layers[-1], rect)
 		self.cam.setCellImageDimensions(*VIEW.CELL_IMAGE_DIMENSIONS)
 		self.cam.setRotation(VIEW.ROTATION)
 		self.cam.setTilt(VIEW.TILT)
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,3 +9,6 @@ select = W2, W3, W6, E11, E125, E13, E211, E223, E224, E225, E226, E227, E228, E
 # https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
 ignore = E101, W191, W503, E241
 max-line-length = 100
+[pytest]
+addopts = --runxfail
+#--runxfail runs tests even if expected to fail.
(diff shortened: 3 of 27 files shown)


squiddy:
That seems to skip pretty much all GUI tests, even the most basic ones.

Looking at the error, my guess is that the tests need a newer version of FIFE, but the package installed in CI is too old (`python-fife`).

ttoocs:
Aye, there's test_trivial(gui), which simply calls pass, and even that fails.
I'm not sure how/why it got into this state, but this would just be a basic step in getting the builds to work again, as it seems every build just fails, and hence some of the easier checks (like the isort and pystyle) and ignored.. and if anything /new/ breaks a test we already have that works, it'd also be ignored -
 entirely defeating the purpose of any of the tests. /shrug

AndyMender:
I think the general approach should be to keep the "master" branch clean (no failing tests) to at least guarantee test-wise that the game will not randomly crash for an average player and address specific issues via topic branches. If broken tests are skipped and Travis CI suddenly starts reporting successful builds, we should at least mark the specific broken tests with a TODO and open a ticket in the Issue tracker to not lose track of them :).

ttoocs:
Agreed on the point of tests. - to verify nothing breaks.
The method of implementation marks each test function with a `@pytest.mark.fixme() `,
However I have not opened up any issues in the issue tracker. If you want me to, I can go make a generic ("Some tests are marked as expected to break, fix them") issue. (Or should there be one issue for each set of tests? Or one for each test..?)

ttoocs:
So, trying to find out where/why everything started breaking, it looks like
broke
One of these broke it initally.. and was merged in here:
broke

And the git get's really messy around there, and I'm unable to run the GUI-tests locally (no idea how/why - SDL just complains of no accelerated gfx.. and I forced it to run on my normal X-display).

So, hopfully if this was fixed, then this pull request would be unneeded.

(i'll keep looking into this and such..)

LinuxDonald:
Yeah Tier5 update have broken thinks. And master branch ist not really stable...

ttoocs:
[the_diff_that_broke.diff.txt](https://github.com/unknown-horizons/unknown-horizons/files/1608607/the_diff_that_broke.diff.txt)

This seems to be the diff between when it worked... and it didn't...

ttoocs:
https://travis-ci.org/ttoocs/unknown-horizons/builds/325738011
^ The same patch applied much closer to the origonal occurance.

AndyMender:
Alright, I'll get to it, run all tests locally and see what's up.

ttoocs:
I was testing it via travis (I am unable to run them locally) 
The viewport change from the last commit does seem to be the _major_ error the breaks all of GUI tests.

However, in testing it more, it seems that even though they're set to 'xfail' which has the ability to have an 'unexpected pass'... they never occur.  (For whatever reason the above build failed on a curl.. but:
https://travis-ci.org/ttoocs/unknown-horizons/builds/325740915 was my run on my fork, ( The exact code as above) (Note, none of the internal tests pass)
While https://travis-ci.org/ttoocs/unknown-horizons/builds/325734499  was a run with just the viewport argument fix applied to master... and `tests/gui/test_internal.py::test_trivial PASSED`

So, I'm aware the major error is just the wrong arguments for viewport.
I'm just confused on why pytest is just skipping all xfail's entirely, not giving the desired Unexpected pass.

AndyMender:
I'm running another round of tests with slightly different settings for Xvfb. Try enabling xinerama and iglx. For me those options allowed using Xvfb properly and not displaying to the main X11 server instance. I too have an nVidia card (Quadro FX 1800) in this rig. The other problem is enet or some errors in the multiplayer part of the game, which show an error window inaccessible through the dummy framebuffer.

test_trivial actually succeeds, but many others don't. I'll pastebin an updated run soon.

ttoocs:
Okay, so I think now, I've got it to work as I wanted:
Tests marked with fixme, are still tested, but don't cause a complete failure. (That way, one can still see if they pass/fail)
While other tests work just normally.

After this build (it should work), I'll remove the fixme flags on the ones that do work, and I think that this is done.

ttoocs:
Or.. pytest will still fail on barracks and traderoute, and I'll be confused again.

AndyMender:
Here is the updated set of tests: https://pastebin.com/hEtsxfxW
and the Shell script I run to execute them: https://pastebin.com/NkjKaqDp
"daemon" is a FreeBSD facility so the /sbin/start-stop-daemon should still be perfectly adequate on Linux. Bare in mind the +iglx and +xinerama options for Xvfb.

ttoocs:
'PS In my conclusion, we should first strive to get an updated version of both FIFE and enet/pyenet on Travis CI.'

Agreed.

PR #2770, opened 2018-01-20
Ai update buildings

LinuxDonald:


LinuxDonald:
@AndyMender maybe we should add public bath here? https://github.com/unknown-horizons/unknown-horizons/blob/ai-update-buildings/horizons/ai/aiplayer/villagebuilder.py

horizons/ai/aiplayer/goal/feederchaingoal.py
```diff
@@ -164,7 +164,7 @@ def get_personality_name(self):
 
 class FeederHygieneGoal(FeederChainGoal):
 	def __init__(self, settlement_manager):
-		super().__init__(settlement_manager, RES.HYGIENE, 'hygiene producer')
+		super(FeederHygieneGoal, self).__init__(settlement_manager, RES.HYGIENE, 'hygiene producer')
```
AndyMender: Is this necessary at all? Per Python 3.x documentation, naming the child object and the binding is no longer needed:
```language="python"
class C(B):
    def method(self, arg):
        super().method(arg)    # This does the same thing as:
                               # super(C, self).method(arg)
```

horizons/ai/aiplayer/goal/productionchaingoal.py
```diff
@@ -215,7 +215,7 @@ def get_personality_name(self):
 
 class HygieneGoal(ProductionChainGoal):
 	def __init__(self, settlement_manager):
-		super().__init__(settlement_manager, RES.HYGIENE, 'hygiene producer')
+		super(HygieneGoal, self).__init__(settlement_manager, RES.HYGIENE, 'hygiene producer')
```
AndyMender: Same as above.

AndyMender:
@LinuxDonald we could add the public bath to villagebuilder.py, but what role would it play then? As a mandatory building with an area of influence which the AI needs to build to cover settler needs? That's how I see it for now :). The positioning might be tricky, though, because it cannot overlap with the other buildings...

I started adding the wooden tower to the game, both as a regular building and for the AI. Which other buildings apart from this one and the hunter do we still require for this PR?

AndyMender:
I'm working on the ai/aiplayer/building/hunter.py code so please don't touch it for the time being :). I'm using ai/aiplayer/fisher.py as a template so that the hunter has some means of calculating wild game availability. @nihathrael I might need your help with reviewing this later :).

AndyMender:
I'm resuming work on this, though I'm not sure how far I'll get.

LinuxDonald:
@jmdejong could you maybe take an look on this pr too? It would be nice to merge this for the next release.

AndyMender:
@LinuxDonald just be aware that this is not finished. I committed what I could, but some AI buildings are missing logic (the hunter, for one).

LinuxDonald:
the ai buildings that not work and be commented out. And maybe @jmdejong could finish it.

jmdejong:
I rebased the branch.
I pushed to [ai-buildings](https://github.com/unknown-horizons/unknown-horizons/tree/ai-buildings) since I'm a bit reluctant to force-push to someone else's branch.

So far it still errors because it's trying to get a nonexistent Producer component template.

LinuxDonald:
Closing this PR the new one is: https://github.com/unknown-horizons/unknown-horizons/pull/2842

PR #2773, opened 2018-01-25
remove RES.HYGIENE

MarkusHackspacher:
At commit 55cc5480c0041b1ecaeb3e298cbd3ca339272598 show travis CI an fault, and this commit remove RES.HYGIENE so is the travis check ok. 

If later the Tier 5 complete, so you could and this.

The that the travis CI is running fine you could look at: 
https://github.com/MarkusHackspacher/unknown-horizons/commits/AIsupportTier5


nihathrael:
This has already been fixed in d66363bbf44b0a1976b6855736f418d3228cb2e3.

PR #2779, opened 2018-03-26
Fix #2717 - tooltips are now on correct side of cursor

### LINKED ISSUE #2717: Tooltip is on wrong side of cursor
For me, the tooltip is on the wrong side of the cursor (right side on the image). Any way to toggle the location? Would be great to know what the building costs and consumes before building it :)

uh_tooltip1

jmdejong:
Tooltips were always shown right of the cursor, even when the cursor was too close to the right edge of the screen.
This is now fixed: when the tooltip would be out of the screen it is shown left of the cursor instead.
This is done by postponing positioning of the tooltip until its size is known.

Please also test this first.
I didn't get any segfaults now, but I didn't get any segfaults either when constantly repositioning the tooltip.
It should not segfault though since the tooltip is only positioned once.

### DIFF
--- a/horizons/gui/widgets/tooltip.py
+++ b/horizons/gui/widgets/tooltip.py
@@ -47,11 +47,9 @@ class _Tooltip:
 		self.bg = None
 		self.label = None
 		self.mapEvents({
-			self.name + '/mouseEntered/tooltip' : self.position_tooltip,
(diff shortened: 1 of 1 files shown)


jmdejong:
This solution is overcomplicated.
Pull request #2780 does the same in a better and simpler way.

PR #2786, opened 2018-04-06
show minimap with same rotation as map

jmdejong:
Since the map is isometric the minimap rotation was always (at least) 45 degrees off.

These changes solve that

Also, the minimap class has been majorly refactored, pulling most of the transformation code in its own class.

### DIFF
--- a/horizons/gui/widgets/minimap.py
+++ b/horizons/gui/widgets/minimap.py
@@ -698,7 +698,7 @@ class Minimap:
 			self.draw()
 
 	def _get_rotation(self):
-		return self.rotation * math.pi / 2
+		return (-self.rotation - 0.5) * math.pi / 2
 
 	def _get_rotated_coords(self, tup):
 		"""Rotates according to current rotation settings.
--- a/horizons/gui/widgets/minimap.py
+++ b/horizons/gui/widgets/minimap.py
@@ -204,8 +204,7 @@ class Minimap:
 		if self.use_rotation:
 			SettingChanged.subscribe(self._on_setting_changed)
 
-		if self.world:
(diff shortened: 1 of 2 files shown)


horizons/gui/widgets/minimap.py
```diff
@@ -51,14 +42,19 @@ def iter_minimap_points(location, world, island_color, water_color, area=None):
 
 	If `area` is set, it's supposed to be a part of `location`, that is to be
 	returned.
+
+	This function is not used anymore in the in-game minimap, but it is necessary
```
AndyMender: Wording - was this function used by the in-game minimap before? If it's used only for the preview, perhaps the code could be moved there?

jmdejong: I wanted to remove this function, and use the _MinimapTransform class for that but that would require some more changes to _MinimapTransform.iter_points.
It's probably still better to fix that first though.

jmdejong:
I just reailized that this PR is still missing some things, mostly related to the "rotate minimap with normal map" setting. When this setting is off it should go back to the old minimap, but the corners are cut of.
I also want to try to make that setting on by default.

horizons/gui/widgets/minimap.py
AndyMender: Then I think it's good as it is right now. I think smaller incremental changes are the way to go :).

AndyMender:
@jmdejong We are grateful to receive any and all contributions. If you feel some extra things should be added to this PR, I'll happily review the additions :).

LinuxDonald:
@jmdejong can this pr get merged or do you want add more stuff to it?

PR #2789, opened 2018-04-23
show minimap correctly without use_rotation

jmdejong:
This was something I wanted to add to te previous pull request.
Without this, only a part of the map is shown if the use_rotation setting is off.
Now the corners of the minimap are shown as well in this situation.


LinuxDonald:
Oh. Thank you for an fast fix :-)

Issue #2450, opened 2016-08-29
Game seems to ignore speed up and speed down options


CrafterSvK:
Clicking + or - doesn't change game speed. Holding doesn't work also.

MasterofJOKers:
Does the Number next to + and - change?

CrafterSvK:
No. But after rebinding to M & N works just fine. Even badly it should have limit.

MasterofJOKers:
So you meant using the + and - keys, doesn't work, not the GUI buttons?

What do you think the limit should be? My old laptop can hardly handle 2x while my newer laptop can handle ~8x and my desktop should be able to handle even more.

MasterofJOKers:
Does + and - work, if you re-bind the functionality to these keys? Did you use +/- from the numpad?

CrafterSvK:
The people are teleporting on 6x speed on my desktop. i3, GTX 650, 8Gb RAM ArchLinux. I used +/- on numpad and in options it was detected. I have no other +/- on my keyboard.

jorgesumle:
Maybe you have number keypad blocked. Is Num Lock on?

LinuxDonald:
I think Max should be 8x or 11x

MasterofJOKers:
Numpad +/- are not necessarily the same +/- as others - even if you don't have other +/-.

CrafterSvK:
Yes it is. I tried with off and it doesn't work anyway. Yeah 8x is okay. More than it should freeze a lot of PCs. I have binded with them so it should work.

MasterofJOKers:
Btw. the current upper bound is 20x.

MasterofJOKers:
Confirmed, that numpad +/- does not work like the other +/-, as they send different keycodes.

Regardingt he cap of speed: I'm against capping the speed, as a human should be able to tell, if its PC is capable of running at this speed or not. Having a hard cap is imho bad, because it is only correct for a small number of PCs.

CrafterSvK:
Why is max cap good. Rapidly tap + for max speed. Don't need to count how many times you want. If you don't want to do that don't do that. But better to do it in options. So you can select uncapped and cap at 8 16 32 and so on.

MasterofJOKers:
Option sound better since "max speed" is different per PC as I already said.

CrafterSvK:
Yeah as I just think of it. :)

jmdejong:
[...]

Issue #2156, opened 2013-08-30
Numpad keys doesnt work ingame


stubb:
If you set up Numpad keys for actions in the hotkey configuration they don't work ingame. Restarting UH doesn't help, so it looks like the keys aren't recognized.
##

LinuxDonald:
Should be fixed with sdl2.

jmdejong:
I believe that this is still a problem (at least for me on archlinux).
The numpad 5 key fires the `NUM_5` key event in the settings menu (just like the number bar 5 key), but it fires the `KP_5` event in the main game view.
`keyevent.getKey.getValue()` actually gives different (integer) keycodes for in the setting menu and the main game.
Also, related to this, the shift key doesn't work in the main game: on my keyboard pressing shift-= gives `PLUS` in the settings menus, but fires the event for `EQUALS` in the main game.

LinuxDonald:
Maybe an problem with fifechan?

helios2000:
Sure,  the related issue here: https://github.com/fifengine/fifengine/issues/981
Code that needs to be adjusted: [Fife](https://github.com/fifengine/fifengine/blob/master/engine/core/gui/fifechan/fifechanmanager.cpp#L401) and [Fifechan](https://github.com/fifengine/fifechan/blob/master/include/fifechan/key.hpp#L137)

PR #2792, opened 2018-04-25
numpad plus and minus keys also set game speed.

jmdejong:
This solves the original problem of issue #2450 by also binding numpad plus and mins keys too speed up and speed down.

These bindings can not be seen or changed in settings menu.
This might be confusing for some players, but it seemed a general improvement anyways.
For some reason the key detector in the settings menu detects numpad plus as `PLUS` instead of `KP_PLUS`, making it impossible to bind the numpad keys from the settings.
Also, there are already two other bindings for speedup and the settings menu can only show 2 bindings.

This only sets the default, so if you already had a settings.xml this won't do anything.
You should delete your settings file (or copy-paste the changed lines from the template) for this to take effect.

### DIFF
--- a/content/settings-template.xml
+++ b/content/settings-template.xml
@@ -83,8 +83,8 @@
 		<Setting name="DESTROY_TOOL" type="list"> X </Setting>
 		<Setting name="ROTATE_RIGHT" type="list"> PERIOD </Setting>
 		<Setting name="ROTATE_LEFT" type="list"> COMMA </Setting>
-		<Setting name="SPEED_UP" type="list"> PLUS ; EQUALS </Setting>
-		<Setting name="SPEED_DOWN" type="list"> MINUS </Setting>
+		<Setting name="SPEED_UP" type="list"> PLUS ; EQUALS ; KP_PLUS</Setting>
+		<Setting name="SPEED_DOWN" type="list"> MINUS ;  ; KP_MINUS</Setting>
 		<Setting name="ZOOM_IN" type="list"> PAGE_DOWN </Setting>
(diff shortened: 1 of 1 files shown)


content/settings-template.xml
```diff
@@ -83,8 +83,8 @@
 		<Setting name="DESTROY_TOOL" type="list"> X </Setting>
 		<Setting name="ROTATE_RIGHT" type="list"> PERIOD </Setting>
 		<Setting name="ROTATE_LEFT" type="list"> COMMA </Setting>
-		<Setting name="SPEED_UP" type="list"> PLUS ; EQUALS </Setting>
-		<Setting name="SPEED_DOWN" type="list"> MINUS </Setting>
+		<Setting name="SPEED_UP" type="list"> PLUS ; EQUALS ; KP_PLUS</Setting>
+		<Setting name="SPEED_DOWN" type="list"> MINUS ;  ; KP_MINUS</Setting>
```
LinuxDonald: is here not an ";" to much?

jmdejong: The second binding for SPEED_DOWN is left empty on purpose.
If the second binding would be used for KP_MINUS then the binding for KP_MINUS would be in the setting menu, but KP_PLUS wouldn't.
I thought it would be more consistent to keep both out of the settings menu.
This is especially because users can not bind KP_MINUS from the settings menu, so if they would change it they could not get it back.

jmdejong:
This whole PR is actually a workaround for a specific case of #2156 by the way (and of the fact that only 2 keybindings can be selected in the settings menu)

PR #2795, opened 2018-04-26
refactor minimap code

jmdejong:
iter_minimap_points now uses other existing methods
get_minimap_color added to decrease code duplication.

This is some cleanup that I originally wanted to include in PR #2786

### DIFF
--- a/horizons/gui/widgets/minimap.py
+++ b/horizons/gui/widgets/minimap.py
@@ -34,70 +34,44 @@ from horizons.messaging import SettingChanged
 from horizons.util.shapes import Circle, Point, Rect
 
 
-def iter_minimap_points(location, world, island_color, water_color, area=None):
+def get_minimap_color(world_coords, world, island_color, water_color):
+
+	full_map = world.full_map
+
+	# check what's at the covered_area
+	if world_coords in full_map:
+		# this pixel is an island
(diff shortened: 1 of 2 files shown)


LinuxDonald:
@AndyMender could you review it?

AndyMender:
@LinuxDonald later today or tomorrow, yes :).

horizons/gui/widgets/minimap.py
```diff
@@ -34,70 +34,44 @@
 from horizons.util.shapes import Circle, Point, Rect
 
 
-def iter_minimap_points(location, world, island_color, water_color, area=None):
+def get_minimap_color(world_coords, world, island_color, water_color):
```
AndyMender: Could you please add a docstring documenting what the function does specifically?
`iter_minimap_points_colors` is nicely documented, but `get_minimap_color `isn't.

jmdejong:
Good point, I added a docstring. Do you think this is good enough or should I add more information?

AndyMender:
@jmdejong in general it's a good practice to use the Sphinx-style docstring blocks for RestructuredText processing (:param:, :return:, etc.). This allows one to build HTML documentation from in-code docstrings with basically one click. However, this project has never been really documented that way so I think your docstring is sufficient :).

PR #2801, opened 2018-05-01
remove unnecessary sleep in startup

jmdejong:
The poll loop will check and sleep when necessary.
No need to perform an extra sleep before.


LinuxDonald:
@AndyMender short review please

AndyMender:
@LinuxDonald It looks alright, but since you already merged it into "master", I can't give my review anymore :).

Issue #2787, opened 2018-04-09
UI objects prevent moving the camera


iluvhalo:
### Intro
I'm sorry if this is a known issue, but it is something I noticed immediately after opening the game. I am on Windows 10, version 1709. The version of the game I am using is the one that the windows setup.exe from the main website sets up. The settings were all default except I changed the resolution to 1366x768 to match my native resolution and changed it to fullscreen. 

### Issue
 After starting the tutorial and looking around the map by moving my cursor to the edge of the screen, I noticed when I put my cursor in the top right corner to move up and right, it didn't move. It seemed like the minimap object was in the way of the boundary that is set up to move the camera. I noticed, too, that when I selected my starting ship, the ship's stat menu popped up and when I tried to pan the camera directly right, the same issue happened. Doing the same thing on the other side of the screen produced similar, but harder to notice results. You can move the camera, but there is a small gap between the objective/goal object ("Build a warehouse...") and the edge of the screen so you can still activate moving the camera, but you can still tell the object gets in the way a little bit. 

### Summary
I don't know how you guys have this code set up, but it definitely seems like the UI objects are getting in the way of the camera movement boundary. If this was a project in GIMP, I would compare it to just having things on the wrong layers. Again, sorry if this is a known issue. I searched the issue tracker for map, UI, and camera and didn't see anything that seemed to describe this. I can tell it's not a finished product, but really great job with everything you guys have done so far.

jmdejong:
I agree. I have encountered this as problem too. I think it's just a matter of setting the size of all the UI widgets.

I do think it's important to fix [issue #2778](https://github.com/unknown-horizons/unknown-horizons/issues/2778) first.
If this issue would be changed first, it would be very likely that you can start moving the viewport by moving the mouse to the edge, but it will keep moving until you get the mouse out of the whole menu area.

jmdejong:
fixed in #2818

Issue #2778, opened 2018-03-26
moving cursor into buildmenu does not end hover actions


jmdejong:
When hovering over a StatusIcon (or is it the building it belongs to?) it shows a tooltip explaining the problem.
This tooltip normally disappears after you move the cursor somewhere else.
However, when you move the cursor directly from the statusicon (or building?) to the build menu, the message stays there until you move the cursor out of the build menu again.

The same problem happens with navigating by moving the cursor to the edge of the screen:
When using the edge of the screen for moving the view, and directly moving the cursor from the edge of the screen into the build menu the view won't stop moving until the cursor is moved out of the build menu again.

To fix this there should probably be some mouseleave that is triggered when the cursor enters the build menu and that does the cleanup

jmdejong:
Another slightly different version of this:
When moving the mouse into the build menu (or some other menu), and then you move the view (with the arrow keys) in such a way that the position where the cursor entered the build menu is over some statusicon, the tooltip is shown too, even though the cursor isn't actually over the building.

jmdejong:
I just found out that part of this problem is already mentioned in #1413 
This is slightly more general because it also mentions the tooltips.

The underlying problem is the same: Fife stops sending mouseMoved events to the main view when the mouse is over some widget, but it does not tell the main view that it will stop firing mouseMoved events or that the mouse has entered some widget.
The main view therefore incorrectly believes that the mouse is at the last place where it was before entering some widget and that the mouse hasn't moved since then

jmdejong:
I just figured out how to make Fife send mouse events when the mouse is over a widget:
Implement a subclass of Fife.IMouseFilter, and implement the isFiltered(self, event) method.
This method will be called before all mouse events on widgets.
If it returns true, the event will still be fired.

jmdejong:
fixed in #2818

Issue #1413, opened 2012-02-07
Scrolling doesn't stop when moving cursor out of the game window


unknown-horizons:
This issue was reported by **AyCe**

When you move the mouse to the left/top/right and then quickly on a UI panel, the scrolling continues (annoying when building)

[> Link to originally reported Trac ticket <](http://trac.unknown-horizons.org/t/ticket/1413)
##

jmdejong:
fixed in #2818

PR #2817, opened 2018-05-10
Fire mousemoved events even when they were consumed by widgets

jmdejong:
Use mousefilter to let the mousemoved and mousereleased events pass through.
This allows the mousetools to properly know the position of the mouse, even when widgets are in front of it.

This PR requires a FIFE to call `MouseEvent.consumedByWidgets()`. See https://github.com/fifengine/fifengine/pull/1059
Do not merge this until FIFE is changed.

This should solve problems #1413, #2778 and #2787.
If this would be done the workaround for #1413 (extending the gui menus) should be undone

### DIFF
--- a/horizons/engine/engine.py
+++ b/horizons/engine/engine.py
@@ -160,7 +160,8 @@ class Fife:
 			# see issue #2778 and #1413
 			# to fix this, mousemove events are fired even when over a widget
 			def isFiltered(self, event):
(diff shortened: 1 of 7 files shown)


jmdejong:
FIFE has gone a slightly different direction, deprecating mousefilters. A slightly different approach will be required.

PR #2818, opened 2018-05-12
NavigationTool listens to mouseevents behind widgets

jmdejong:
All its subclasses as well.
This is done by setting it as a global listener.
This requires the latest git version of FIFE.
isConsumedByWidgets is used to check whether to fire or not.

This should solve problems #1413, #2778 and #2787.

This probably requires some playtesting.
It is likely that I missed some case where the mouseevent should not have fired through a widget.


AndyMender:
@jmdejong the changes don't seem too radical and code-wise they look clean. If @LinuxDonald is okay with them, this can be merged.

LinuxDonald:
@AndyMender broke UH :D

PR #2821, opened 2018-05-13
fixed and ran generate_credits script

jmdejong:
The generate credits script still had some python2-only stuff.
I changed that and ran the script to update the actual credits.


AndyMender:
Perhaps it would be better to implement "generate_credits" as a sub-component of UH, rather than a standalone script? That way the credits would be generated automatically every time the game is launched, for instance from a static YAML or JSON.

jmdejong:
We could do that. We can make it just like the atlases that they run on startup in development mode.
Using YAML or JSON will definitely make parsing easier than the mardown that's being parsed now.

On the other hand, the credits don't change very often and there is not much to test/debug about the credits. We should just make sure to run the script before the releases. I'd say it's not very urgent right now.

AndyMender:
In that case we can implement it as part of a "pre-release" build process before launching a release.

PR #2825, opened 2018-05-13
Change the layout of the minimap

jmdejong:
minimaplayout2
Since the minimap is now rotated, the layout of the minimap area should be somewhat different too.


horizons/gui/widgets/minimap.py
```diff
@@ -279,18 +281,64 @@ def update_cam(self):
...
-		for i in range(0, 4):
-			self.minimap_image.rendertarget.addLine(self._get_render_name("cam"),
-			                                        minimap_corners_as_point[i],
-			                                        minimap_corners_as_point[(i + 1) % 4],
-			                                                         *self.COLORS["cam"])
+		xmin, ymin = self.transform.world_to_minimap(displayed_area[0])
+		xmax, ymax = self.transform.world_to_minimap(displayed_area[2])
+
+		# draw horizontal lines
+		for y in [ymin, ymax]:
```
AndyMender: Tuples are lighter than lists so it's slightly better to use a tuple here:
```python
for y in (ymin, ymax):
```

horizons/gui/widgets/minimap.py
```diff
@@ -279,18 +281,64 @@ def update_cam(self):
...
+			x1 = max(xmin, self.transform.get_min_x(y))
+			x2 = min(xmax, self.transform.get_max_x(y))
+			if x2 > x1:
+				drawLine(fife.Point(x1, y), fife.Point(x2, y))
+
+		# draw vertical lines
+		for x in [xmin, xmax]:
```
AndyMender: Same as for the earlier comment.

horizons/gui/widgets/minimap.py
```diff
@@ -753,14 +760,32 @@ def iter_points(self):
 				x = x_i + location.left
 				# TODO: better name
 				asdf = int(abs(x / location.width - 0.5) * location.height)
-				for y_i in range(asdf, location.height - asdf):
+				for y_i in range(self.get_min_y(x), self.get_max_y(x)):
```
AndyMender: Good change. The earlier code wasn't very clear. You can remove `asdf ` entirely now, I think.

jmdejong: You're right! I didn't even realize.

horizons/gui/widgets/minimap.py
jmdejong: I thought I deleted this code in a later commit?

AndyMender: My bad, yes, you did.

PR #2832, opened 2018-05-16
crop unit images to adjust origin.

### LINKED ISSUE #2824: Image Origins
All images have a single point from which they're drawn.
Since #2804 this is the center of the ground area of the image, at location (width/2, height-width/4).
tile_drawing
The red dot is the origin; the lines surround the ground area.

In order for all entities to show up on the correct position and z-ordering, this origin should match correctly.
For buildings this means that the ground area should match. For trees this means that the place where the trunk meets the ground should be the origin. For units the origin should be between their feet.

For most images the origin follows this pattern, but there are a few exceptions:
- The most notable is for deers. Because they are too far above the image origin, they will appear in front of trees that should be in front of them. Also, when looking at them from different angles they might actually appear at different positions.
- Humans are slightly below the origin in the image. The effect is that they never seem to walk in the center of the street, but just below it.
- Ground tiles tried to simulate elevation. See #2807

This can be fixed by adjusting the empty space around all incorrect images.
Alternatively all images could have an optional metadata file that describes the origin.

jmdejong:
Messed around with imagemagick.
Ships are not adjusted for now.
Also, sailor women are still not really correctly positioned. For some reason they are shorter than other units.

This should hopefully fix #2824.


LinuxDonald:
I will test this tomorrow and merge it than.

PR #2836, opened 2018-05-17
fix test_settler_level_save_load

jmdejong:
The test_settler_level_save_load test was supposed to test if a house upgrades correctly after saving and reloading the session.
It failed to take into account that there might not be enough inhabitants or that the inhabitants aren't happy enough.
This sets the taxes so that the inhabitants stay happy, and it makes sure that there are enought inhabitants.

The test still doesn't really make sense since there are 2 sessions: the one passed as argument and a newly created session. The building is placed in the argument session, and the waiting is there as well, while the save/load is done on the new session. I tried to fix this but that crashed it.
At least the test now makes a bit more sense.

### DIFF
--- a/horizons/world/building/settler.py
+++ b/horizons/world/building/settler.py
@@ -220,13 +220,9 @@ class Settler(BuildableRect, BuildingResourceHandler, BasicBuilding):
 
 	def _tick(self):
 		"""Here we collect the functions, that are called regularly (every "month")."""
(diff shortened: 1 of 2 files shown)


AndyMender:
I'm not sure whether it's the same test, but one of the tests had an issue with civilization upgrading after save & load when moving from level 4 to level 5. The one you're talking about is a different one?

PR #2844, opened 2018-05-20
Follow XDG basedir specifications; fix #2749

### LINKED ISSUE #2749: Follow XDG basedir specification
Right now a new folder for configurations and userdata is created in `$HOME/.unknown-horizons`.
To be more flexible it would be nice to implement the [XDG Base Directory Specification][1].

> `$XDG_DATA_HOME` defines the base directory relative to which user specific data files should be stored. If `$XDG_DATA_HOME` is either not set or empty, a default equal to `$HOME/.local/share` should be used.
> `$XDG_CONFIG_HOME` defines the base directory relative to which user specific configuration files should be stored. If `$XDG_CONFIG_HOME` is either not set or empty, a default equal to `$HOME/.config` should be used.
> `$XDG_CACHE_HOME` defines the base directory relative to which user specific non-essential data files should be stored. If `$XDG_CACHE_HOME` is either not set or empty, a default equal to `$HOME/.cache` should be used.

An easy explanation can be found in the [gnome wiki][2].

This would result in the following folders:

- `$XDG_CONFIG_HOME/unknown-horizons/<SETTINGS.XML>` defaulting to `$HOME/.config/unknown-horizons/<SETTINGS.XML>` when `$XDG_CONFIG_HOME` is not set or empty.
- `$XDG_DATA_HOME/unknown-horizons/<LOG|MAPS|SAVE|SCREENSHOTS>` defaulting to `$HOME/.local/share/unknown-horizons/<LOG|MAPS|SAVE|SCREENSHOTS>` when `$XDG_DATA_HOME` is not set or empty.
- `$XDG_CACHE_HOME/unknown-horizons/<YAMLDATA.CACHE>` defaulting to `$HOME/.cache/unknown-horizons/<YAMLDATA.CACHE>` when `$XDG_CACHE_HOME` is not set or empty.

Advantages are:
- No predefined paths. The user can decide where his or her data is located.
- No cluttered home directory. Some selectors will automatically show dotfiles, finding the files you want will be easier.
- A shared storage location for data. Application data does not have to be searched for, as these are no longer located in different places. This also benefits a backup.
- Resetting the settings and keeping the application data at the same time becomes easier, as there is a clear separation.

[1]: https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
[2]: https://wiki.gnome.org/Initiatives/GnomeGoals/XDGConfigFolders#Guidelines

jmdejong:
Unknown Horizons now puts user files in `$XDG_CONFIG_HOME/unknown-horizons`, `$XDG_DATA_HOME/unknown-horizons` and `$XDG_CACHE_HOME/unknown-horizons`.

For linux, the defaults are `~/.config/unknown-horizons`, `~/.local/share/unknown-horizons` and `~/.cache/unknown-horizons`.
For windows, the only difference is that the cache is placed in it's own directory (though it will also follow the XDG environment variables if they exist)

You can also override the paths using uh environment variables.
`$UH_USER_DIR` sets the directory for all user files. Data and config are both in this directory, and cache is in it's own subdirectory.
`$UH_USER_CONFIG_DIR`, `$UH_USER_DATA_DIR` and `$UH_USER_CACHE_DIR` can be used to set these directories indivually. These more specific options override `$UH_USER_DIR`.

If the new directories do not exist but the old ones do, the files at the old directory will be automatically moved to the new directory.
For config and data a symlink will be created from the old location to the new one, so users will be able to find their files back. This is also useful when switching back and forth between UH versions. A note describing the migration is left in the data directory. This note is also printed to stdout when the migration is happening. The note links to the wiki for more information: https://github.com/unknown-horizons/unknown-horizons/wiki/2018-user-data-migration

Users aren't informed about cache migration, because they shouldn't need to look at it manually anyways and when switching versions cache is very likely to be incompatible anyways

What do you think of this? I chose to make symlinks and leave a note instead of showing a popup because the migration only matters when users are looking at the directories anyways.

Also, could you also double-check for typos in filenames and environment variables? These are things that aren't easily caught. Maybe this part needs some tests?

### DIFF
--- a/horizons/util/migratepaths.py
+++ b/horizons/util/migratepaths.py
@@ -1,3 +1,23 @@
+# ###################################################
+# Copyright (C) 2018 The Unknown Horizons Team
+# team@unknown-horizons.org
+# This file is part of Unknown Horizons.
+#
+# Unknown Horizons is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the
+# Free Software Foundation, Inc.,
+# 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+# ###################################################
 
 import glob
 import os
--- a/horizons/util/migratepaths.py
+++ b/horizons/util/migratepaths.py
@@ -1,0 +1,79 @@
+
+import glob
+import os
+
+from horizons.constants import OLDPATHS, PATHS
+
+
+def migrate_paths():
+
+	migrated = False
+	if check_cache_migration():
+		migrate_cache()
+		# cache migration doesn't really need a link or mention I think
+
+	if check_config_migration():
+		migrate_config()
+		link_config()
+		migrated = True
+
+	if check_data_migration():
+		migrate_data()
+		link_data()
+		migrated = True
+
+	if migrated:
+		leave_note()
+
+
+def check_cache_migration():
+	return os.path.isdir(OLDPATHS.CACHE_DIR) and not os.path.isdir(PATHS.CACHE_DIR)
+
+
+def migrate_cache():
+	cache_files = glob.glob(os.path.join(OLDPATHS.CACHE_DIR, "*.cache"))
+	for oldname in cache_files:
+		newname = os.path.join(PATHS.CACHE_DIR, os.path.basename(oldname))
+		os.renames(oldname, newname)
+
+
+def check_config_migration():
+	return os.path.isfile(OLDPATHS.USER_CONFIG_FILE) and not os.path.isfile(PATHS.USER_CONFIG_FILE)
+
+
+def migrate_config():
+	os.renames(OLDPATHS.USER_CONFIG_FILE, PATHS.USER_CONFIG_FILE)
+
+
+def link_config():
+	os.symlink(PATHS.USER_CONFIG_FILE, OLDPATHS.USER_CONFIG_FILE)
+
+
+def check_data_migration():
+	return os.path.isdir(OLDPATHS.USER_DATA_DIR) and not os.path.isdir(PATHS.USER_DATA_DIR)
+
+
+def migrate_data():
+	os.renames(OLDPATHS.USER_DATA_DIR, PATHS.USER_DATA_DIR)
+
+
+def link_data():
+	os.symlink(PATHS.USER_DATA_DIR, OLDPATHS.USER_DATA_DIR)
+
+
+def leave_note():
+	notefile = os.path.join(PATHS.USER_DATA_DIR, "migration_readme.txt")
+	notetext = """
+In the latest version of unknown-horizons the user data directories have been moved.
+This has been done automatically, and symlinks have been created to minimize inconvenience.
+
+It is safe to remove these symlinks and this readme file.
+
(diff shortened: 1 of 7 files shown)


horizons/util/migratepaths.py
```diff
@@ -0,0 +1,79 @@
...
+In the latest version of unknown-horizons the user data directories have been moved.
+This has been done automatically, and symlinks have been created to minimize inconvenience.
+
+It is safe to remove these symlinks and this readme file.
+
+If you ever downgrade the UH version you might need more care.
+
+For more information, see the wiki:
+https://github.com/unknown-horizons/unknown-horizons/wiki/2018-user-data-migration
+"""
```
baka0815: Maybe add an information about the new directories so the user can find the new files?

horizons/util/platform.py
```diff
@@ -31,21 +30,86 @@ def get_home_directory():
...
 	else:
-		return home_directory.joinpath('My Games', 'unknown-horizons')
+		return os.path.join(home_directory, 'My Games', UH)
+
+
+def _try_directories(paths, default=None):
+	# takes a list of lists of strings as first argument
+	# returns the joined path for the first list where all strings exist and are nonempty
+	# or the second argument there is no such list
+	# (lists coul also be tuples of course)
```
baka0815: Typos:
Or the seco n argument *IF* there is...

(lists coul*D* also...)

horizons/util/platform.py
```diff
@@ -31,21 +30,86 @@ def get_home_directory():
...
+	Returns a triplet of directories for game-related data.
+	The first value is the directory for configuration files (settings.xml)
+	The second is the directory for game data (saves, maps, screenshots, logs)
+	The third is the directory for cache (atlas-metadata.cache, yamldata.cache)
+	"""
+	home_directory = get_home_directory()
+
+	# these are the default directories
+	if platform.system() == "Windows":
+		user_dir = os.path.join(home_directory, 'My Games', UH)
```
baka0815: In case of Windows the user_dir is set, otherwise it isn't. Is this wanted?

jmdejong: user_dir is only a helper variable in this function. It's scope is this function and it is not returned. It is just to make the next 3 assignments a bit shorter.

horizons/util/migratepaths.py
jmdejong: Yes, I should still improve that wiki page.
The symlinks should allow the user to find the files initally. I prefer linking to the wiki as much as possible (since we can edit it without update), but maybe it's indeed better to put the actual filepaths in there.

jmdejong:
https://docs.python.org/3/library/os.html#os.symlink
Apperently windows supports symlinks from Vista onwards. Since XP isn't supported by microsoft anymore I think it's safe to use symlinks

jmdejong:
What should I do with the licence by the way? should I just copy the licence from another file and change the year? should I keep 2008 as starting year or 2018?

horizons/util/migratepaths.py
baka0815: Just use the variables, than it's no provlem if the directories change. And it's a one time change...

horizons/util/platform.py
baka0815: I just found it a bit confusing, that there was a difference between Win and non-Win.

jmdejong: There is always a difference between windows and non-windows because windows doesn't have `.cache`, `.local/share` and `.config` directories. That's why all uh user data is sill in the same directory (I haven't found a much better place to put it). It made sense to me to put this common directory in a helper variable.

AndyMender:
@jmdejong 2008 is the official starting date of the project, though as far as I know, the license should have an up-to-date date, so 2018. @LinuxDonald correct me if I'm wrong here :).

LinuxDonald:
For an new file the date must be that on in that year that it got created. You dont need to back date it.

AndyMender:
@jmdejong the changes look good in general. I was initially a bit skeptical about splitting directories, but I see that we actually need that so it's fine :).

PR #2857, opened 2018-06-28
make configure-traderoute button inactive if there are no destinations

### LINKED ISSUE #2721: Don't offer route configuration with only one settlement
If the player only has one settlement, it is still possible to open the route configuration menu, only to find out that one has to select at least two settlements.

Shouldn't we just disable this feature until two settlements were build?

florian-s-code:
Fix #2721 

Next step is to make the button look inactive. The xml template has a random image instead of a real inactive image so it may not be very clear.
Should I try to improve on that in the same Pull Request ?


LinuxDonald:
@jmdejong or @AndyMender review please.

LinuxDonald:
@yhoyhoj could you maybe do the change to if ... else ... as @jmdejong have written it please?

LinuxDonald:
Thank you for your pr :)

Issue #2859, opened 2018-07-02
Ships display lags behind real position or is totally wrong (master branch)


yhoyhoj:
Master branch on Archlinux.  
After ordering a ship to move, it will often stay in place on the display while its position on the map changes.
The ship can then apparently keep an offset for some time. It can end up being displayed on the islands.  
  
How to reproduce: start a game, move ship.
   
Screenshot 1: Ship position on the display is not the same as on the map. I believe the map is right and that it is the main display that is wrong.
![Uploading 2018-07-02.18-48-33.120988.png…]()  
Screenshot 2: AI ship is on the island. It is not clear but I think the map shows it right next to the island.
![2018-07-02 18-52-49 651280](https://user-images.githubusercontent.com/2287301/42176518-84c1b7ac-7e29-11e8-8901-572a39e856d8.png)

LinuxDonald:
yeah the minimap needs to be fixed.

jmdejong:
Weird, I've never seen this and I can't reproduce it (with the same setup).

Does anything change when you rotate, move or zoom the main map view?
Are atlases enabled?
Does UH print anything to the console when you start a game?
Does it happen in all maps you play? If not, could you share your save?
Does it also happen when you start uh with the command line option `--start-random-map`?

@LinuxDonald What should be fixed about the minimap? Maybe the ship icons could be drawn a little higher, but I don't think that's the problem here.

LinuxDonald:
Could we not scale down all the information from the main map and load the information on the minimap?

yhoyhoj:
No output, except : "Ship created and weapons loaded." when the game starts.
After some further investigation, it doesn't happen all the time.
It doesn't with `--start-random-map` but it does when starting a random map from the menu.
It doesn't when playing a scenario (the_Unknown or tutorial).
It does in Free Play.
  
Zooming or moving the ship out of view and then in will not change anything. The ship keeps its glitched position until it is moved.
I am not sure but it looks like the ship sprite will sometimes play catch-up with the real position and the offset will vary during a move (the ship trajectory on the minimap and the sprite trajectory will not be the same, in this case, but they meet. Not always when the ship reach the position it is supposed to move to, though).
  
All ships are affected  : player, AI player, free trader and I suppose pirate, but I could not verify.

yhoyhoj:
[...]

PR #2863, opened 2018-07-29
sync graphics location for moving objects

jmdejong:
Units have 2 different pathfinders.
One is the UH one in python, which deals with the long path, but only works in integer coordinates.
The other one is Fife's route system, which does the part between two paths on UH's pathfinder so the movement looks smooth.

For some unknown reason, Fife's route system is not always able to go where it is supposed to go and the ship will get stuck. It then tries to correct with absurd movements. See #2859.

This patch will make sure that the location of the fife instance is never too far from the actual location by checking the fife location whenever the actual location updates.
If the two locations are too far away from each other, the fife location will be set to the actual location.

This might result in a little jump now and then, but those are the cases where the ship would have otherwise been stuck.
If this error is rare enough it won't be noticable.
Also, the error was more common on the higher game speeds where a little jump won't be noticable anyways

### DIFF
--- a/horizons/world/units/movingobject.py
+++ b/horizons/world/units/movingobject.py
@@ -209,6 +209,18 @@ class MovingObject(ComponentHolder, ConcreteObject):
 		else:
 			self.__is_moving = True
 
+
+		# HACK: 2 different pathfinding systems are being used here and they
+		# might not always match.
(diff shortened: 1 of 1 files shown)


LinuxDonald:
@helios2000 what do you think about this issue?

Issue #2839, opened 2018-05-17
editor brush resets when moving cursor over brush selection widget


jmdejong:
If you reset the cursor to the normal cursor in the editor (by right clicking), then move the cursor over the brush selector widget (without clicking anything) and then back to the main view, the cursor turns into the tilelayer brush again by itself.
This makes it somewhat harder to click buttons in another menu (like the 'save map' button).

Rootyjr:
I think I fixed this issue, though I'm still waiting on the pull request.

LinuxDonald:
Should be fixed with: https://github.com/unknown-horizons/unknown-horizons/pull/2864

PR #2864, opened 2018-07-30
Fixed the tile bug. See #2839

Rootyjr:
Fixed the tile bug.  See #2839


LinuxDonald:
@jmdejong have you time for an review please?

PR #2867, opened 2018-08-13
Fix test collection under python3.7

asottile:
All I did was:

1. create a python3.7 virtualenv
2. install a few dependencies and `pytest` in the virtualenv
3. `pytest tests --collect-only` (and fix the error)

### DIFF
--- a/tests/dummy.py
+++ b/tests/dummy.py
@@ -44,7 +44,7 @@ class Dummy:
 			if attr in ('__base__', '__bases__', '__basicsize__',
 				'__dictoffset__', '__flags__', '__itemsize__',
 				'__members__', '__methods__', '__mro__', '__name__',
-				'__subclasses__', '__weakrefoffset__',
+				'__subclasses__', '__weakrefoffset__', '__mro_entries__',
 				'_getAttributeNames', 'mro'):
 				raise
 			else:


asottile:
👋 @ttoocs

LinuxDonald:
Does it works too with py 3.6?

asottile:
yep yep, `__mro_entries__` is a new feature in python3.7 for making the `typing` module faster.  not proxying this attribute in python3.6 shouldn't change anything (nothing uses it)

LinuxDonald:
We use typing too in our game code. Could that maybe speed up it too? And do you maybe would like to work on it?

asottile:
most of the time savings in the `typing` module is import-time (which might help with startup here? not super familiar with the project myself).  There's some other cool perf improvements in python3.7 like fastcall and [others](https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-perf)

The unfortunate news is I actually know very little about this project.  This PR is entirely the result of an IRC conversation in the pytest irc channel (#pylib) on freenode where I helped with some debugging (I help maintain pytest).  So while I'm always happy to help and I'd love to contribute, I'm not sure I have the time commitment or proper motivation to do it justice.

PR #2871, opened 2018-09-01
Fix Travis fail with isort

MarkusHackspacher:
add the libraries to the config file


LinuxDonald:
Sorry for the long delay.

PR #2881, opened 2018-12-15
Also create parent directories for game files

jmdejong:
When the paths for the game directories (data, config and cache) do not exist UH now does not only automatically create these directories themselves, but also their parent directories.

The drawback of this is that if some user misspelled one of the paths they probably won't notice it because UH will just make a directory with the misspelled path.
A second opinion on what is most desirable would be nice.


jmdejong:
I just realized that this is kind of a follow-up to #2870.

I think we should use this. If someone would manually set the game directories they'll probably know what they're doing anyways.

PR #2910, opened 2019-07-18
replace deprecated to distro package

MarkusHackspacher:
platform.dist()[0].lower() will be deprecated beginning with Python 3.7.
replace with distro.linux_distribution(full_distribution_name=False)[0]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 coverage==4.4.1
 coveralls==1.1
 crayons==0.1.2
+distro
 future==0.16.0
 greenlet==0.4.15
 isort==4.3.4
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,6 +1,7 @@
 coverage==4.4.1
 coveralls==1.1
 crayons==0.1.2
+distro
```
pauricthelodger: Might be worth pinning it to 1.4.0 since that's the current release in https://pypi.org/project/distro/?

MarkusHackspacher: It is better if the version is not fixed, as the distros are constantly evolving.

Normally should someboby continuous pinned Packages for a new version, test the new version und tell the provider if the package have a problem.

PR #2932, opened 2019-12-04
Fix mistranslation

AMDmi3:


LinuxDonald:
Thanks for the fix. But translations must be fixed here: https://hosted.weblate.org/projects/uh/

PR #2940, opened 2020-03-28
Updated all Finnish translations

lahtis:
Updated all Finnish translations in 100%. Using Weblate. And later I found projects dead.


LinuxDonald:
Translation we will port to our new godot version that is in work

LinuxDonald:
I have merged the translations from weblate. The auto script is broken....

PR #2944, opened 2020-07-03
generate_atlases: dont fail deleting files

bmwiedemann:
dont fail deleting files in `generate_atlases.py` if the files are already gone.

Without this patch, build on 1-core VMs sometimes failed with
```
 intltool-merge -d po/uh/ content/packages/unknown-horizons.desktop.in build/share/applications/unknown-horizons.desktop
 Merging translations into build/share/applications/unknown-horizons.desktop.
 + python3 horizons/engine/generate_atlases.py 2048
 Traceback (most recent call last):
   File "horizons/engine/generate_atlases.py", line 427, in <module>
     AtlasGenerator.clear_everything()
   File "horizons/engine/generate_atlases.py", line 407, in clear_everything
     os.unlink(path)
 FileNotFoundError: [Errno 2] No such file or directory: '/home/abuild/.cache/unknown-horizons/atlas-metadata.cache'
```

This PR was done while working on reproducible builds for openSUSE.

### DIFF
--- a/horizons/engine/generate_atlases.py
+++ b/horizons/engine/generate_atlases.py
@@ -401,10 +401,13 @@ class AtlasGenerator:
 
 		# delete everything
 		for path in paths:
-			if not os.path.exists(path):
-				continue
 			cls.log.info('Deleting %s', path)
-			os.unlink(path)
+			try:
+				os.unlink(path)
+			except FileNotFoundError:
+				# file was already gone
+				# and that is the state we wanted anyway
+				pass
 
 
 if __name__ == '__main__':


horizons/engine/generate_atlases.py
```diff
@@ -404,7 +404,10 @@ def clear_everything(cls):
 			if not os.path.exists(path):
```
jmdejong, line 404: I think this if should be removed too since the try ... except does the same

horizons/engine/generate_atlases.py
```diff
@@ -404,7 +404,10 @@ def clear_everything(cls):
...
 				continue
 			cls.log.info('Deleting %s', path)
-			os.unlink(path)
+			try:
+				os.unlink(path)
+			except FileNotFoundError:
+				pass
```
jmdejong, line 410: I've always learned that you should never have an empty catch statement.

I can see that it is reasonable here, but I think it would be better to at least log it

bmwiedemann: The previous "continue" was not logging it either.

bmwiedemann: Now that the other check is dropped, there might be more log entries from here.

jmdejong: That's fair. Seeing this empty block feels wrong to me, but it's probably the best indeed.
Maybe add a comment that mentions what you just said: that this is the desirable state anyways

Issue #2954, opened 2021-02-08
[BUG] TypeError: __init__() got an unexpected keyword argument 'encoding'


Lukas1818:
Direct after the start of the game I get a type TypeError. A window does apper for a short time, but it does not show anything.

OS: Manajro Linux (based on Arch)
[unknown horizons AUR-package](https://aur.archlinux.org/packages/unknown-horizons/) 2019.1-4
[fife AUR-package](https://aur.archlinux.org/packages/fife/) 0.4.2-3 with [fix](https://github.com/fifengine/fifengine/pull/1078/commits/cf295fd98a8fba080f6305c27be56d10ab7ce94d) for the [deprecated getchildren()](https://docs.python.org/3.9/whatsnew/3.9.html#removed)
```
$ unknown-horizons
[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)
[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)
Traceback (most recent call last):
  File "/usr/bin/unknown-horizons", line 381, in <module>
    main()
  File "/usr/bin/unknown-horizons", line 122, in main
    ret = horizons.main.start(options)
  File "/usr/lib/python3.9/site-packages/horizons/main.py", line 179, in start
    Entities.load(horizons.globals.db, load_now=False) # create all references
  File "/usr/lib/python3.9/site-packages/horizons/entities.py", line 64, in load
    cls.load_grounds(db, load_now)
  File "/usr/lib/python3.9/site-packages/horizons/entities.py", line 77, in load_grounds
    tile_sets = TileSetLoader.get_sets()
  File "/usr/lib/python3.9/site-packages/horizons/util/loaders/tilesetloader.py", line 72, in get_sets
    cls.load()
  File "/usr/lib/python3.9/site-packages/horizons/util/loaders/tilesetloader.py", line 65, in load
    cls.tile_sets = JsonDecoder.load(PATHS.TILE_SETS_JSON_FILE)
  File "/usr/lib/python3.9/site-packages/horizons/util/loaders/jsondecoder.py", line 40, in load
    return json.load(f, encoding="ascii", object_hook=_decode_dict)
  File "/usr/lib/python3.9/json/__init__.py", line 293, in load
    return loads(fp.read(),
  File "/usr/lib/python3.9/json/__init__.py", line 359, in loads
    return cls(**kw).decode(s)
[...]

PR #2955, opened 2021-02-09
fixs for Phyton 3.9

Lukas1818:
* fix TypeError, see #2954
* The isAlive() function is deprecated and was change to is_alive().
>The isAlive() method of threading.Thread has been removed. It was deprecated since Python 3.8. Use is_alive() instead. (Contributed by Dong-hee Na in bpo-37804.)

https://docs.python.org/3.9/whatsnew/3.9.html#removed


LinuxDonald:
@jmdejong review please

jmdejong (approved):
Good work!

PR #2962, opened 2023-08-13
Python Pillow 10 compatibility

CrafterSvK:
Game cannot be run on system with Pillow 10 installed because of unknown Image contant.
Image.ANTIALIAS was deprecated and removed from Pillow, now it uses only name LANCZOS which was used from Pillow version 5.0


CrafterSvK:
`requirements.txt` states `Pillow>=7.1.0` which means by installing dependencies you would still install version 10.0

